PR #4, opened 2021-02-10
ADAPT-000: Add Identity Bar, Logo, Global Footer, SR only Text, Button

yvonnetangsu:
# READY

# Summary
- Add Logo, Identity Bar, Global Footer, Button, SrOnlyText components

# Needed By (Date)

# Urgency
- How critical is this PR?

# Steps to Test

1. Pull down to local, nvm use, npm run dev
1. Look at Storybook local site and check that the new components are there
1. Look at code

- JIRA ticket
- Other PRs

# See Also


package.json
```diff
@@ -85,7 +78,8 @@
     "react-hero-icon": "^0.0.7",
     "react-scripts": "3.4.4",
     "storybook-addon-designs": "^5.4.3",
-    "storybook-dark-mode": "^1.0.4"
+    "storybook-dark-mode": "^1.0.4",
+    "tailwindcss-classnames": "^2.0.1"
```
sherakama: I went with just `classnames` as the benefits of using it requires you to configure it with our custom tailwind.config.js file. You won't get any of our custom stuff without doing that.

yvonnetangsu: Yeah I got rid of that package too - I think this was an old commit.

src/Button/Button.js
```diff
@@ -1,24 +1,27 @@
...
+import { buttonStyles, buttonSizes } from './Button.levers'
 
 /**
- * Primary UI component for user interaction
+ * Button Component
+ *
+ * HTML button element
  */
-export const Button = ({ className, children, onClick, type, ref, ...props }) => {
+export const Button = ({ classes, children, onClick, style, size, ref, ...props }) => {
```
sherakama, line 10: I have to say, it felt more natural to write `className` on this component but it does break a bit from the convention on the alert nested object. We should have a convention for this. My rough thinking was that one level components like button, link, could use className, whereas the more complicated components could use the large object.

src/Button/Button.js
```diff
@@ -29,12 +32,14 @@ export const Button = ({ className, children, onClick, type, ref, ...props }) =>
 Button.propTypes = {
   // HTML Button type.
   type: propTypes.oneOf(['button', 'reset', 'submit']),
```
sherakama, line 34: The type property should be treated as a lever as well.

src/Button/Button.js
```diff
@@ -29,12 +32,14 @@ export const Button = ({ className, children, onClick, type, ref, ...props }) =>
...
+  style: propTypes.oneOf(['primary', 'secondary', 'disabled']),
```
sherakama, line 35:
```suggestion
  style: propTypes.oneOf(buttonStyles),
```

src/Button/Button.js
```diff
@@ -29,12 +32,14 @@ export const Button = ({ className, children, onClick, type, ref, ...props }) =>
...
+  size: propTypes.oneOf(['default', 'big', 'small']),
```
sherakama, line 36:
```suggestion
  size: propTypes.oneOf(buttonSizes),
```

src/Button/Button.js
yvonnetangsu: Yeah, I'm ok with that

src/Button/Button.js
```diff
@@ -13,12 +16,50 @@ export const Button = ({ className, children, onClick, type, ref, ...props }) =>
...
+      case 'minimal':
+        levers.size = classnames('su-p-0')
+        break
+    }
+  }
+
   return (
     <button
-      className={classnames('su-button', levers.button, className)}
+      className={classnames('su-button', levers.variant, levers.size, className)}
```
sherakama, line 59:
```suggestion
      className={classnames('su-button', ...levers, className)}
```

You should only have to include the object as classnames should be able to handle that.

sherakama:
Some inspiration for the button component https://github.com/rcaferati/react-awesome-button#awesomebutton-props

sherakama:
And more: https://github.com/arifszn/reactive-button#available-props

src/Button/Button.js
yvonnetangsu: Hmm I tried this but it didn't seem to work

src/Logo/Logo.js
```diff
@@ -0,0 +1,81 @@
...
+      case 'short':
+        logoText = 'Stanford';
+        break;
+
+      case 'full':
+        logoText = 'Stanford University';
+        break;
+
+      case 'stacked':
+        logoText = 'Stanford \n University';
```
yvonnetangsu, line 46: @sherakama You might have a quick answer to this - what's the best way for me to pass in "Stanford<br />University" as HTML here?

sherakama: Perhaps something like this?

```suggestion
        logoText = (<>Stanford <br /> University</>);
```

yvonnetangsu: Yay! That works - thank you!

sherakama:
> @sherakama I think this is ready for you to take a look - all the test passed and it build locally. I looked at the build error on Netlify but I'm not sure how to fix it and get it to build 🤔

https://github.com/facebook/jest/issues/6880

Seems to be a bug. Let me see if we can work around this.

src/Button/Button.js
```diff
@@ -1,54 +1,109 @@
...
-  const classnames = require('classnames')
-  const levers = {}
+  const classnames = require('classnames');
+  const levers = {};
 
   // Levers
   // ---------------------------------------------------------------------------
 
+  // Props.variant & if button not disabled
+  if (variant && buttonVariants.includes(variant) && !isDisabled) {
```
sherakama, line 20:
```suggestion
  // Props.variant 
  if (variant && buttonVariants.includes(variant)) {
```

I don't think we need to check for disabled. You can have a disabled secondary button no?

yvonnetangsu: I'm going to ask Kerri if we can have all the disabled button with the same style since the design for the disabled secondary is not accessible.

src/Button/Button.stories.js
```diff
@@ -1,17 +1,84 @@
...
+      }
+    },
+    type: {
+      control: {
+        type: 'select',
+        options: buttonTypes
+      }
+    },
+    onClick: {
+      action: 'clicked'
```
sherakama, line 29: nice

src/Button/Button.js
sherakama: Proposed some changes in a new PR. Check it out.

src/GlobalFooter/GlobalFooter.js
```diff
@@ -0,0 +1,145 @@
...
+        levers.wrapper = classnames('su-bg-black');
+        break;
+    }
+  }
+
+  return (
+    <div className={classnames('su-global-footer su-body-basefont-20 su-rs-py-1 su-text-white su-link-white hover:su-link-white focus:su-link-white', levers.wrapper, className)}>
+      <div className='su-cc su-flex su-flex-col lg:su-flex-row' title='Common Stanford resources'>
+        <div className='su-text-center su-mt-5 su-mb-9'>
+          <Logo className={classnames('su-type-3')} color={levers.logo} type="stacked" />
```
sherakama, line 37:
```suggestion
          <Logo className={classnames('su-type-3')} type="stacked" />
```

levers.logo is never set.

yvonnetangsu: Good catch - it's the default color so doesn't need to pass a color.

src/IdentityBar/IdentityBar.stories.js
```diff
@@ -0,0 +1,53 @@
...
+      control: {
+        type: 'select',
+        options: IdentityBarColors
+      }
+    },
+  }
+};
+
+// Set up an Alert Template.
+const IdentityBarTemplate = ({ ...rest }) => <IdentityBar {...rest} />;
```
sherakama, line 23:
```suggestion
const IdentityBarTemplate = (props) => <IdentityBar {...props} />;
```

src/Logo/Logo.stories.js
```diff
@@ -0,0 +1,61 @@
...
+      control: {
+        type: 'select',
+        options: LogoTypes
+      }
+    },
+  }
+};
+
+// Set up an Alert Template.
+const LogoTemplate = ({ children, ...rest }) => <Logo {...rest} />;
```
sherakama, line 25:
```suggestion
const LogoTemplate = (props) => <Logo {...props} />;
```

yvonnetangsu:
@sherakama Is this ready to merge you think?

sherakama (approved):
GTG. Nice work. Thanks for all the feedback/revisions.

yvonnetangsu:
> GTG. Nice work. Thanks for all the feedback/revisions.

Yay! Thank YOU for the very helpful feedback/suggestions.

PR #6, opened 2021-02-17
Minor fixes and suggestions.

sherakama:
# READY FOR REVIEW


src/Button/Button.js
```diff
@@ -1,56 +1,55 @@
...
+import PropTypes from 'prop-types';
+import { buttonVariants, buttonSizes, buttonTypes } from './Button.levers';
 
 /**
  * Button Component
  *
  * HTML button element
  */
-export const Button = ({ className, children, onClick, ref, ...props }) => {
+export const Button = ({ className, children, onClick, ref, variant, size, type, isDisabled, ...props }) => {
```
sherakama, line 10: If you destructure each of the levers in the constructor you can keep the `{...props}` on the button HTML element which might come in handy. If you don't destructure the props arg then you can't use `{...props}` on the HTML element as React doesn't know what to do with props like `isDisabled: bool`

src/Button/Button.js
```diff
@@ -1,56 +1,55 @@
...
+        break;
 
       case 'minimal':
-        levers.size = classnames('su-p-0')
-        break
+        levers.size = classnames('su-p-0');
+        break;
+
+      default:
+        levers.size = classnames('su-px-26 su-py-10 su-text-20');
```
sherakama, line 52: Moved default to the default so you don't have to pass in the keyword `default`.

src/Button/Button.js
```diff
@@ -101,4 +100,4 @@ Button.defaultProps = {
   size: 'default',
```
sherakama, line 100:
```suggestion
```

src/Button/Button.levers.js
```diff
@@ -1,14 +1,14 @@
...
  * Styles for the button.
  */
-export const buttonVariants = ['primary', 'secondary', 'none']
+export const buttonVariants = ['primary', 'secondary', 'none'];
 
 /**
  * Size of the button
  */
-export const buttonSizes = ['default', 'big', 'small', 'minimal']
+export const buttonSizes = ['default', 'big', 'small', 'minimal'];
```
sherakama, line 9:
```suggestion
export const buttonSizes = ['big', 'small', 'minimal'];
```

yvonnetangsu (approved):
Cool - thanks!

PR #7, opened 2021-02-17
ADAPT-000: More preview stories.

sherakama:
# READY FOR REVIEW 

# Summary
- Adds a few more preview stories of basic HTML elements

# Needed By (Date)
- Whenever

# Urgency
- None

# Steps to Test

1. Review build
2. Review code
3. Make comments on whether or not these are useful.


.storybook/stories/Video.stories.js
```diff
@@ -0,0 +1,73 @@
...
+
+Video.propTypes = {
+  className: PropTypes.string,
+  autoplay: PropTypes.oneOf(['true', 'false']),
+  loop: PropTypes.oneOf(['true', 'false']),
+  muted: PropTypes.oneOf(['true', 'false']),
+  preload: PropTypes.oneOf(['auto', 'metadata', 'none']),
+  controls: PropTypes.oneOf(['true', 'false']),
+  poster: PropTypes.string
+}
```
yvonnetangsu, line 64: Is there a reason to use oneOf instead of boolean? And looks like there's a mix of true/false in the args with or without single quotes in these.

yvonnetangsu:
Thanks for adding these - looks great, just that one question about the true/false stuff

.storybook/stories/Audio.stories.js
```diff
@@ -0,0 +1,59 @@
...
+
+export const Controls = Template.bind({})
+Controls.args = {
+  className: '',
+  controls: 'true',
+  autoplay: false,
+  loop: 'false',
+  muted: 'false',
+  preload: 'none',
+}
```
yvonnetangsu, line 30: Booleans in audio too please 😄

sherakama: Oh, duh!

.storybook/stories/Audio.stories.js
```diff
@@ -0,0 +1,59 @@
...
+  controls: PropTypes.bool,
+}
+
+Audio.defaultProps = {
+  autoplay: false,
+  loop: false,
+  muted: false,
+  preload: 'none',
+  controls: true,
+  src: 'https://interactive-examples.mdn.mozilla.net/media/cc0-audio/t-rex-roar.mp3',
```
yvonnetangsu, line 58: Nice choice 🦖

sherakama: rawr

sherakama:
@yvonnetangsu 

This is ready for review again. I removed the Jest testing tab from the UI as it is causing problems and is added overhead we don't need now that we run them on PRs through github actions.

.storybook/stories/Video.stories.js
```diff
@@ -0,0 +1,73 @@
...
+
+export const NoControls = Template.bind({})
+NoControls.args = {
+  className: '',
+  controls: false,
+  autoplay: true,
+  loop: false,
+  muted: false,
+  preload: 'none',
+}
```
yvonnetangsu, line 43: Looks like if doing it this way and use the ...props cause the video/audio to go away completely if the controls are set to false.

sherakama: https://stackoverflow.com/questions/9565363/video-element-disappears-in-chrome-when-not-using-controls

Seems to be a bug in chrome

yvonnetangsu: Yikes 🤦🏼‍♀️

yvonnetangsu: Looks like you set some other attributes to true which works better. I'm looking at the video with no controls here (the ocean one) and that one works too, so maybe it needs one of those attributes set to true.
https://decanter.stanford.edu/component/simple-media/

yvonnetangsu (approved):
Thanks for adding all these elements! GTG. Feel free to merge when you're ready - in case you want to do more with the video thing.

PR #12, opened 2021-02-19
ADAPT-000: Bundlers and ESModule Classnames alternative

sherakama:
# READY FOR REVIEW

# Summary
- Upgrades packages
- Swaps `classnames` for `clsx` as classnames was causing issues with bundlers and tests
- Adds `npm run package` to compile new packages
- Removes broken Jest tests for the time being

# Needed By (Date)
- End of sprint

# Urgency
- Med/low
- 
# Steps to Test

1. Check out storybook preview build
2. Review code
3. Checkout branch locally and run `nvm use && npm install`
3. Pull into your latest work in https://github.com/SU-SWS/ood_online_giving/pull/11 and rebuild
4. Ensure nothing breaks in https://github.com/SU-SWS/ood_online_giving/pull/11 after running `npm update && npm run develop`
5. Run jest tests `npm run test`
6. Run `npm run build`
7. Run `npm run dev`
8. Run `npm run package`


src/Logo/Logo.js
```diff
@@ -43,13 +43,13 @@ export const Logo = ({ className, ...props }) => {
         break;
 
       case 'stacked':
-        logoText = (<>Stanford<br />University</>);
+        logoText = (<>Stanford <br /> University</>);
```
sherakama, line 46: Fragments should work again.

package.json
```diff
@@ -37,32 +50,39 @@
...
-    "@storybook/node-logger": "^6.1.15",
-    "@storybook/react": "^6.1.15",
+    "@storybook/addon-viewport": "^6.1.19",
+    "@storybook/node-logger": "^6.1.19",
+    "@storybook/react": "^6.1.19",
     "@testing-library/jest-dom": "^5.11.9",
     "@testing-library/react": "^11.2.5",
     "@whitespace/storybook-addon-html": "^4.2.0",
     "babel-jest": "^26.6.3",
-    "classnames": "^2.2.6",
```
yvonnetangsu, line 54: Uh oh, does that mean no dedupe? 👀

sherakama: https://www.npmjs.com/package/clsx#usage

dedupe is the default behavior of this package. It is more modern and quicker than classnames. Also, https://github.com/JedWatson/classnames/issues/220

yvonnetangsu: Yeah pretty cool. I'm checking it out. It's really popular too. Nice find.

sherakama: Ah, disabled button needs to be re-worked... arg.

.babelrc
```diff
@@ -0,0 +1,14 @@
...
+    "@babel/preset-env",
+    "@babel/preset-react"
+  ],
+  "plugins": [
+    "@babel/plugin-transform-react-jsx",
+    ["transform-jsx-classnames",
+      {
+        "dedupe": true
+      }
+    ]
```
yvonnetangsu, line 12: Sounds like we don't need this anymore. Yay!

yvonnetangsu:
@sherakama You probably know this already, but the HTMl elements under `.storybook/stories` look like they need to be tweaked also. They are lower priority though since those are just for demo.

.babelrc
sherakama: Yeah, can probably pull that out now

sherakama:
Yeah, let me re-work the clsx library a bit. It looks like it isn't a 100% replacement for classnames and only falsy values are dropped instead of removed. I'll look at the dedupe code in classnames and repurpose it a bit with a wrapper around clsx to restore it to the former glory.

yvonnetangsu:
@sherakama I can confirm that it works after I did `npm update` and `npm run develop/build` in the Online Giving repo 🎉 

In this Decanter-React repo - build, dev, and package all work for me with some issues that we've already discussed.

sherakama:
FYI: https://github.com/lukeed/clsx/issues/13

sherakama:
@yvonnetangsu This is ready to review again. I'll update https://github.com/SU-SWS/ood_online_giving/pull/11 with the latest for easy review. 

Please also review https://github.com/SU-SWS/clsx-dedupe (index.js). That has the dedupe code re-written in a modern style and will export as an ESModule.

I've noticed a build Warning message when running `npm run build` now: `WARN unable to find package.json for clsx-dedupe` which has me boggled right now. It doesn't seem to bother anything but is something I'd like to fix.

Please have another review.

src/Alert/Alert.levers.js
```diff
@@ -1,4 +1,4 @@
-const classnames = require('classnames')
+import clsx from 'clsx';
```
yvonnetangsu, line 1: Are we using `clsxd` for everything or just as needed?

sherakama: I would say just as needed. No sense on adding cpu cycles.

sherakama: Probably can remove those from the levers.

sherakama:
Raw editing through github UI FTW.

yvonnetangsu:
@sherakama I'm looking at the Storybook preview build, and everything looks to be working as intended except the HTML elements.
^^^ (Moved this comment to the right repo/branch)

src/Alert/Alert.levers.js
yvonnetangsu: Thank you for that whole bunch of revisions!

yvonnetangsu:
I pulled down this #rollup branch locally, did nvm use, npm install, npm run build, npm run test, npm run dev, npm run package.

I noticed the warning you mention above when using `npm run build`, and the other problem being the HTML elements not working at the moment. Other than that, everything works great! Oh, looks like you just pushed up another commit. Let me pull it.

sherakama:
Good catch. That should fix the HTML elements.

sherakama:
Man, we need purgecss on this package too!

yvonnetangsu (approved):
All working! Thank you for the great work 🎉  😄

sherakama:
Ok, merging down. Don't forget to review the `clsx-dedupe` repo too! Send in a PR with any comments or questions you may have to that project.

yvonnetangsu:
`clsx-dedupe` looks good! Learned some JS features that are new to me like `arg.flat()` while looking at your code 😄

sherakama:
> `clsx-dedupe` looks good! Learned some JS features that are new to me like `arg.flat()` while looking at your code 😄

Yay code review!

PR #15, opened 2021-02-27
Adapt-1955 Adapt-1700 Adapt 1960 Logo Lockup, Local Footer, Heading components

yvonnetangsu:
# READY FOR REVIEW

# Summary
- New components: Lockup, Local Footer, Heading
- Tweaks to FlexCell and GridCell. FlexCell can now specify width for each breakpoint; GridCell can now specific column span for each breakpoint.
- Add unlink option to Stanford Logo component

# Needed By (Date)

# Urgency
- How critical is this PR?

# Steps to Test

1. Look at preview build and check the example stories and play with the controls
1. Look at code

- JIRA ticket
- Other PRs

# See Also


src/FlexCell/FlexCell.js
```diff
@@ -21,13 +21,64 @@ export const FlexCell = ({ className, children, ref, ...props }) => {
...
+  // props.xs
+  if (props.xs && flexCellWidth.includes(props.xs)) {
+    if (typeof props.xs === 'string') {
+      levers.xs = `su-w-${props.xs}`;
     }
     else {
-      levers.width = `su-w-${props.width}/12`;
+      levers.xs = `su-w-${props.xs}/12`;
+    }
+  }
```
sherakama, line 33: Hmm, this could lead to a little bit of a DX issue. 

`<FlexCell xs="9" />` Wouldn't set the xs breakpoint and it would not produce an error. It would quietly not add anything. Enforcing the integer vs the string value of a number is desired but I would expect this component to throw an error if the wrong input type is passed. Also, if this value was being delivered from an API, often, it will be sent as a string instead of an integer.

src/Grid/Grid.stories.js
```diff
@@ -94,10 +94,10 @@ const GridTemplate = ({ children, ...rest }) => {
 const GridPageTemplate = ({ ...rest }) => {
   return (
     <Grid {...rest}>
-      <GridCell element='header' colSpan={12} className='su-bg-sky su-p-20'>Header</GridCell>
-      <GridCell element='aside' colSpan={12} className='md:su-col-span-4 su-bg-plum-light su-p-20'>Sidebar</GridCell>
-      <GridCell element='main' colSpan={12}  className='su-row-start-2 md:su-row-start-auto md:su-col-span-8 su-bg-plum su-p-20 su-h-200'>Main</GridCell>
-      <GridCell element='footer' colSpan={12} className='su-bg-palo-verde su-p-20'>Footer</GridCell>
+      <GridCell element='header' xs={12} className='su-bg-sky su-p-20'>Header</GridCell>
+      <GridCell element='aside' xs={12} md={4} className='su-bg-plum-light su-p-20'>Sidebar</GridCell>
+      <GridCell element='main' xs={12} md={8}  className='su-row-start-2 md:su-row-start-auto su-bg-plum su-p-20 su-h-200'>Main</GridCell>
```
sherakama, line 99:
```suggestion
      <GridCell element='main' xs={12} md={8} className='su-row-start-2 md:su-row-start-auto su-bg-plum su-p-20 su-h-200'>Main</GridCell>
```

src/FlexCell/FlexCell.js
yvonnetangsu: Oh interesting...that's a good point.

yvonnetangsu: Sounds like perhaps the best approach is for me to make it work for both cases - whether an integer or a string is passed?

src/GridCell/GridCell.js
```diff
@@ -21,13 +21,65 @@ export const GridCell = ({ className, children, ref, ...props }) => {
...
+
+  // props.xxl
+  if (props.xxl && gridColSpan.includes(props.xxl)) {
+    if (props.xxl === 'auto') {
+      levers.xxl = `2xl:su-col-auto`;
+    }
+    else {
+      levers.xxl = `2xl:su-col-span-${props.xxl}`;
     }
   }
```
sherakama, line 84: This whole thing could probably be re-written in a loop to avoid code duplication.

eg:
```
const bps = ['xs', 'sm', 'md', 'xl', '2xl'];
bps.forEach((bp) => {
  if (props[bp] && gridColSpan.includes(props[bp]) {
	if(props[bp] === "auto") {
	  levers[bp] = bp + ':su-col-auto';
	}
	else {
	  levers[bp] = bp + ':su-col-span-' + props[bp];
	}
  }
}
```

src/GridCell/GridCell.js
```diff
@@ -37,13 +89,13 @@ export const GridCell = ({ className, children, ref, ...props }) => {
...
+  if (props.row && gridRowSpan.includes(props.row)) {
+    levers.row = `su-row-span-${props.row}`;
+    if (props.row === 'auto') {
+      levers.row = `su-row-auto`;
     }
     else {
-      levers.rowSpan = `su-row-span-${props.rowSpan}`;
+      levers.row = `su-row-span-${props.row}`;
     }
   }
```
sherakama, line 100: No breakpoints for row?

src/GridCell/GridCell.js
yvonnetangsu: That's cleaner :)

src/GridCell/GridCell.js
yvonnetangsu: I feel like we have lots of props already, and rowSpan is used rarely so I'm thinking that if people really need to use responsive rowSpan they could add it to the classname 🤔

src/FlexCell/FlexCell.js
sherakama: I'd expect one of two things to happen.
1. Throw an error if there is invalid input 
2. Output my invalid input and let the developer troubleshoot it later. 

By not outputting anything quietly, the component makes troubleshooting much more difficult. I'll leave it up to you on which direction you would like to go with it. Throwing errors or accepting loosely typed parameters is a valid solution to each. If we were writing TypeScript the strict typing of the parameter would be required and you'd likely have to throw the error.

src/GridCell/GridCell.js
sherakama: Ok, leave a code comment to that extent. I can imagine someone a couple years down the line wondering how to do this for their edge case.

src/Heading/Heading.js
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * Heading Component
+ *
+ */
+export const Heading = ({ className, level = 3, font, weight, tracking, align, size, uppercase, italic, srOnly, children, ref, ...props }) => {
+  // Defaults & Variables.
+  // ---------------------------------------------------------------------------
+  const levers = {};
+  let HeadingTag = '';
```
sherakama, line 14: Should this default to `h2`?

yvonnetangsu: Sure. `h2` it is

src/LocalFooter/LocalFooter.js
```diff
@@ -0,0 +1,169 @@
...
+
+  return (
+    <Container width='site' className={clsxd('su-local-footer su-rs-pt-4 su-rs-pb-5', levers.wrapper, classes.wrapper)}>
+      <FlexBox direction='col' gap={true} className={clsxd('md:su-flex-row su-rs-mb-2', classes.header)} justifyContent={'space-between'}>
+        <FlexCell grow={true}>
+          <Lockup color={levers.lockup} line1={props.line1} url={props.url} isLink={props.isLogoLink} />
+        </FlexCell>
+        {props.hasButton && (
+          <FlexCell grow={false}>
+            <a href={props.buttonUrl} className={clsxd('su-inline-block su-bg-digital-red su-text-white su-text-18 hocus:su-text-white hocus:su-bg-archway-dark su-py-9 su-px-20 su-no-underline su-font-regular hover:su-underline focus:su-underline su-rounded-7 su-shadow-md', classes.button)}>
```
sherakama, line 68: Missing an option for href. 

Ideally, this would be refactored to use a button/link component and would be passed in as a prop or child.

sherakama (changes requested):
Feedback complete.

src/LocalFooter/LocalFooter.js
yvonnetangsu: Yeah, was thinking that we'll be using a subcomponent after the CTA link component is done.

yvonnetangsu: > Feedback complete.

Thank you for going through this huge PR and the helpful feedback - I'm working on the revision 😃

src/FlexCell/FlexCell.js
yvonnetangsu: I was experimenting with errors, and saw that PropTypes did throw an error in the console if I tried to use a string
Screen Shot 2021-03-08 at 5 30 02 PM

Is this too quiet still? Do you think perhaps I could add a classname `su-w-ERROR` for example or is that a bad idea?

src/GridCell/GridCell.js
yvonnetangsu: Do you happen to know what's the proper syntax for doing this for destructured props? E.g., in Grid component, the props xs, sm etc are all destructured so I tried the below and a few other permutations but couldn't get it to work (couldn't get `bp` to dynamically reference the destructured data with the same name as the breakpoint instead of a string) and my google fu was failing me 😬 

  ```
bps.forEach((bp) => {
      if ([bp] && gridNumCols.includes([bp])) {
        levers[bp] = `su-grid-cols-${[bp]}`;
      }
    }
  );
```

Comment (yvonnetangsu) [reply src/GridCell/GridCell.js]: I tried this and it seems to work. Is this ok?

```
let bps = {xs: xs, sm: sm, md: md, lg: lg, xl: xl, xxl: xxl};

  Object.entries(bps).forEach(([key, value]) => {
    if (value && gridNumCols.includes(value)) {
      if (key === 'xs') {
        levers[key] = `su-grid-cols-${value}`;
      }
      else if (key === 'xxl') {
        levers[key] = `2xl:su-grid-cols-${value}`;
      }
      else {
        levers[key] = `${key}:su-grid-cols-${value}`;
      }
    }
  });
```

File: src/Heading/Heading.js [hunk 7]
```diff
@@ -0,0 +1,133 @@
...
+
+  // Levers
+  // ---------------------------------------------------------------------------
+
+  // props.level - allows number or string input
+  if (level && (headingLevels.includes(level) || headingLevels.includes(Number(level)))) {
+    HeadingTag = 'h' + level;
+  }
```
yvonnetangsu, line 22: Trying the approach to make it work for both number and string input. Are you ok with this? 😬

sherakama: Good enough for me.

src/Heading/Heading.js
sherakama:
```suggestion
  let HeadingTag = 'h2';
```

src/Heading/Heading.js
```diff
@@ -0,0 +1,133 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { headingLevels, headingFonts, headingAlign, headingWeights, headingTracking, headingSizes } from "./Heading.levers";
+import clsxd from 'clsx-dedupe';
+
+/**
+ * Heading Component
+ *
+ */
+export const Heading = ({ className, level = 2, font, weight, tracking, align, size, uppercase, italic, srOnly, children, ref, ...props }) => {
```
sherakama, line 10:
```suggestion
export const Heading = ({ className, level, font, weight, tracking, align, size, uppercase, italic, srOnly, children, ref, ...props }) => {
```

src/Heading/Heading.js
```diff
@@ -0,0 +1,133 @@
...
+  // props.italic
+  if (italic) {
+    levers.italic = `su-italic`;
+  }
+
+  // props.srOnly
+  if (srOnly) {
+    levers.srOnly = `su-sr-only`;
+  }
+
```
sherakama, line 63:
```suggestion
```

yvonnetangsu:
> Ok, thanks for this work. Looking good.

Thanks! I'll clean up some more in another PR, but glad that this is approved and can be merged first. Let me npm run package it first.

src/GridCell/GridCell.js
yvonnetangsu: Yeah I went with what worked at the time - I got the syntax wrong somewhere when I tried other approaches. Will ask you later when you have time.

yvonnetangsu:
Hmm @sherakama, I merged, but found that the preview site on Netlify is not pulling the latest Decanter v7 updates (I used the #v7 branch for package.json) even though on my local it does pull in those classes. 
For example, on https://decanter-react.netlify.app/?path=/story/stanford-identity-local-footer--button (lockup and button don't look right)
Screen Shot 2021-03-09 at 10 36 51 AM

On my local (looks right)
Screen Shot 2021-03-09 at 10 37 38 AM

I already did a `rm -rf node_modules`, and `rm package-lock.json`, then `npm install` again before I packaged and pushed up the final commit. Wonder what's causing this 🤔 ?

yvonnetangsu:
Maybe I need to run dist on Decanter v7. Hmmm....

sherakama:
I can take some time for branches to come through on npmjs.org.

PR #18, opened 2021-03-16
ADAPT-1701 CTA Link and Link Buttons

yvonnetangsu:
# READY FOR REVIEW

# Summary
- Add CTA link (button and non-button) component

# Needed By (Date)

# Urgency
- How critical is this PR?

# Steps to Test

1. Go to preview build and look at the CTA Link and CTA Link Button examples
1. Look at code

- JIRA ticket
- Other PRs

# See Also


src/CtaLink/CtaLink.js
```diff
@@ -0,0 +1,268 @@
...
+
+      case 'white':
+        levers.color = 'su-text-white hocus:su-text-white';
+        break;
+    }
+  }
+
+  // isButton
+  if (isButton) {
+    levers.isButton = 'su-cta-btn su-font-regular su-leading-none';
```
sherakama, line 55: This lever name is a bit misleading. Perhaps change it to `levers.wrapper`

src/CtaLink/CtaLink.js
```diff
@@ -0,0 +1,268 @@
...
+
+      case 'block':
+        levers.display = 'su-block';
+        break;
+
+      case 'flex':
+        levers.display = 'su-flex';
+        break;
+    }
+  }
```
sherakama, line 34: Do you need a display prop if classname is available and only one css class is needed?

src/CtaLink/CtaLink.js
```diff
@@ -0,0 +1,268 @@
...
+
+  // color
+  if (color && ctaLinkColors.includes(color) && !isButton) {
```
sherakama, line 37: Why not allow for buttons to have color?

src/CtaLink/CtaLink.js
```diff
@@ -0,0 +1,268 @@
...
+  // isButton
+  if (isButton) {
+    levers.isButton = 'su-cta-btn su-font-regular su-leading-none';
+  }
+  else {
+    levers.isButton = 'su-cta-link su-text-19 md:su-text-20';
+  }
+
+  // variant
+  if (variant && ctaLinkButtonVariants.includes(variant) && isButton) {
```
sherakama, line 62: So the variant prop only works if the link is a button? I'd like to see props not rely or depend on another prop setting in order to work. This could lead to bad DX. What happens if you allow for links and buttons to have variants?

src/CtaLink/CtaLink.js
```diff
@@ -0,0 +1,268 @@
...
+        break;
+
+      case 'ghost':
+        levers.variant = 'su-bg-transparent su-text-white hocus:su-text-white su-border-2 su-border-white su-border-solid';
+        break;
+    }
+  }
+
+  // size
+  if (size && ctaLinkButtonSizes.includes(size) && isButton) {
```
sherakama, line 79: Perhaps the button link style should be its own component? There are quite a few conditionals on  the isButton prop.

src/CtaLink/CtaLink.js
```diff
@@ -0,0 +1,268 @@
...
+      case 'top-right':
+        levers.animate += 'group-hocus:su-translate-x-01em group-hocus:su--translate-y-01em';
+        break;
+
+      case 'down':
+        levers.animate += 'group-hocus:su-translate-y-02em';
+        break;
+
+      case 'none':
+        break;
```
sherakama, line 168: This is not necessary and can be removed.

src/CtaLink/CtaLink.js
```diff
@@ -0,0 +1,268 @@
...
+    }
+  }
+
+  // animate
+
+  // Common classes if animation is needed
+  if ((animate && ctaLinkAnimations.includes(animate)) && animate !== 'none') {
```
sherakama, line 148: I think the `none` option can be removed.

src/CtaLink/CtaLink.js
```diff
@@ -0,0 +1,268 @@
...
+    levers.animate = 'su-transition-transform group-hocus:su-transform ';
+  }
+
+  // Add specific classes for each type of animation
+  if (animate && ctaLinkAnimations.includes(animate)) {
+    switch(animate) {
+      case 'right':
+        levers.animate += 'group-hocus:su-translate-x-02em';
```
sherakama, line 156: instead of `+=` concatenation use `clsx`

src/CtaLink/CtaLink.js
```diff
@@ -0,0 +1,268 @@
...
+        break;
+
+      case 'none':
+        break;
+    }
+  }
+
+  return (
+    <a
+      className={clsxd(levers.isButton, 'su-w-fit su-no-underline hover:su-underline focus:su-underline su-group su-transition-colors',
```
sherakama, line 174: I'd like to see the default stuff first in the `clsxd` so that they can be undone later. Why is the lever for button first?

src/CtaLink/CtaLink.js
```diff
@@ -0,0 +1,268 @@
...
+      ref={ref}
+      {...props}
+    >
+      {text}
+      {srText &&
+        <span className='su-sr-only'> {srText}</span>
+      }
+      {icon &&
+        <Icon icon={heroicon} type='solid' aria-hidden={true} className={clsxd('su-inline-block', levers.icon, levers.animate, classes.icon)}/>
+      }
```
sherakama, line 186: It would be nice if we would allow the developer to pass in their own icon and provide this one as a default.

src/CtaLink/CtaLink.levers.js
```diff
@@ -0,0 +1,38 @@
+// Levers shared by both CTA Links and CTA Link Button components.
+// ---------------------------------------------------------------------------
+
+/**
+ * Display type
+ */
+export const ctaLinkDisplay = ['inline-block', 'block', 'flex'];
```
sherakama, line 7: This might be prop overkill. This can be sent in using className

src/CtaLink/CtaLink.levers.js
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Icon options
+ */
+export const ctaLinkIcons = ['action', 'more', 'external', 'lock', 'download', 'video', 'jump', 'email', 'none'];
+
+/**
+ * Icon animation options
+ */
+export const ctaLinkAnimations = ['right', 'top-right', 'down', 'none'];
```
sherakama, line 17:
```suggestion
export const ctaLinkAnimations = ['right', 'top-right', 'down'];
```

sherakama (changes requested):
A number of comments left. They deserve more discussion and thought. We can chat about them or continue the discussion here.

src/CtaLink/CtaLink.js
yvonnetangsu: levers.wrapper sounds good

src/CtaLink/CtaLink.js
yvonnetangsu: Yeah probably not needed. I'll likely add a CTA group component later and can control the display through that parent component.

src/CtaLink/CtaLink.js
yvonnetangsu: This is just the color of the text. We tend to not change button text color much (since it's either white, or digital red - for the outline one, or totally custom by passing in classes).

src/CtaLink/CtaLink.levers.js
yvonnetangsu: I'll get rid of the display prop

src/CtaLink/CtaLink.levers.js
yvonnetangsu: I do find it useful to have a "none" option - since in Storybook, when you're using the controls to switch between options, it won't let you have an "unselected" or unset state once you pick something. So having this allow the user to unset the selection. I agree that if it's not because of Storybook, I'd just leave it out.

src/CtaLink/CtaLink.js
yvonnetangsu: Yeah, this is in the plans 😄

src/CtaLink/CtaLink.js
yvonnetangsu: Sure, I'll put the default stuff first. In this current version, the component class name goes into the isButton lever so I put it first, but it doesn't need to be.

src/CtaLink/CtaLink.js
yvonnetangsu: Please see answer below and let me know what you think.

src/CtaLink/CtaLink.js
yvonnetangsu: It would work if I remove this - it is not needed at all, just wondered if having this would make it clearer (so will have all the options in the lever map to its own case) so more like a documentation thing. If you think it's better not to have it, I'll be happy to remove.

src/CtaLink/CtaLink.js
yvonnetangsu: 😂  I started out having the button link its own component, then I realized that there were quite a bit of shared code (the icons and animation stuff), so I decided to combine them. I did make 2 separate "stories" files though so it's a bit more manageable.

Want to make a decision?

src/CtaLink/CtaLink.js
yvonnetangsu: This is related to the answer right below this one - let me know what you think.

yvonnetangsu:
> A number of comments left. They deserve more discussion and thought. We can chat about them or continue the discussion here.

Thank you for the quick review! I'm on the fence about some of these approaches myself - let me know what you think.

src/CtaLink/CtaLink.js
yvonnetangsu: This is gone

src/CtaLink/CtaLink.js
yvonnetangsu: This has been removed.

src/CtaLink/CtaLink.js
yvonnetangsu: This has been revised

src/CtaLink/CtaLink.js
yvonnetangsu: Done.

src/CtaLink/CtaLink.js
yvonnetangsu: Done

src/CtaLink/CtaLink.js
yvonnetangsu: I'm thinking about adding a "none" or "unset" option to the button variants, so it's easier when someone want to pass custom TW classes for colors and they won't get overridden by the color classes we added for the button variants.

And, would it help if I rename this option `buttonVariants` or `buttonStyles`? I know that's still not ideal - let me know if you rather we have 2 separate files for links and link buttons.

src/CtaLink/CtaLink.js
sherakama: > 😂 I started out having the button link its own component, then I realized that there were quite a bit of shared code (the icons and animation stuff), so I decided to combine them. I did make 2 separate "stories" files though so it's a bit more manageable.
> 
> Want to make a decision?

Abstraction, and the right amount, can be tough. If you have shared items amongst multiple objects you can pull them out into a higher parent if you are using inheritance but alas we are not. We tend to import shared libraries in most of our work. Do a little reading (I'll do the same) at different abstraction techniques in Javascript/React, and we can come back to the design on this.

jbickar: Yvonne and I talked about this higher-level issue in our 1:1 yesterday. One of the fundamental challenges that we face in SWS is writing code that scales; i.e., abstraction. Thanks to you both for putting the thought and discussion into this.

yvonnetangsu: Yeah agree about finding the right balance. I'll poke around and see if I can find a good example of sharing common options.

yvonnetangsu: After reading that article you shared, and thinking about decanter 6 experience, I think I'm going to separate out CTA Link and CTA Link Button into 2 different components so it doesn't have a crap load of props, and it will be easier for future maintenance. I might or might not experiment with sharing some of the levers code (eg, icons) in this PR. Let me know what you think.

sherakama: I think that is a good approach. Every article I read argues for composition over configuration and the more we can do that in this repo the better. We are knowingly going against recommended practices in order to support our design team and would do good by ourselves to try to keep to some best practices where we can. More components and shared levers might give us a chance to grow and maintain this thing before it becomes too complicated and untestable.

yvonnetangsu: Sounds good 😄  I'll try my best to keep it simple stupid.

sherakama:
@yvonnetangsu close this in favor of https://github.com/SU-SWS/decanter-react/pull/20?

yvonnetangsu:
Yes good to close. Thanks!

PR #20, opened 2021-03-24
Adapt 1701 cta v2

yvonnetangsu:
# READY FOR REVIEW

# Summary
- This is based on the other CTA branch, but with separate CTA Button and CTA Link component and some shared levers/functions to reduce redundancy.

# Needed By (Date)

# Urgency
- How critical is this PR?

# Steps to Test

1. Do this
1. Then this
2. Then this

- JIRA ticket
- Other PRs

# See Also


src/CtaButton/CtaButton.js
```diff
@@ -0,0 +1,153 @@
...
+    heroicon = getIconOptions(icon);
+    levers.icon = getIconClasses(icon);
+  }
+
+  // animate
+
+  // Add specific classes for each type of animation
+  if (animate && iconAnimations.includes(animate)) {
+    levers.animate = getIconAnimation(animate);
+  }
```
yvonnetangsu, line 70: @sherakama Trying new approach. Is it worth doing this you think?

src/common/icon/icon.levers.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * CTA Icon options
+ */
+export const iconOptions = ['action', 'more', 'external', 'lock', 'download', 'video', 'jump', 'email', 'none'];
+
+/**
+ * CTA Icon animation options
+ */
+export const iconAnimations = ['right', 'top-right', 'down', 'none'];
```
yvonnetangsu, line 9: Shared icon levers

src/common/icon/getIconAnimation.js
```diff
@@ -0,0 +1,28 @@
+import clsxd from 'clsx-dedupe';
+
+const getIconAnimation = (animate) => {
```
yvonnetangsu, line 3: One of the shared functions that returns a value that is dependent on one of the icon levers..

src/CtaButton/CtaButton.js
yvonnetangsu: Since we're separating out the CTA Link and Link Button components, we could either duplicate the icon/animation code and have them in each file, or have them shared this way. The duplicate way is more readable, but this way has less redundant code (there might be a better way to structure this).

yvonnetangsu: Also, more levers/values are sharable this way, e.g., the button sizes and variants are sharable with the CTA button links.

sherakama:
I'll have a look tomorrow.

yvonnetangsu:
> I'll have a look tomorrow.

Sleep time!

src/Button/Button.js
```diff
@@ -33,76 +44,119 @@ export const Button = ({ className, children, onClick, ref, variant, size, type,
...
 
-  // Props.size
+  // size
   if (size && buttonSizes.includes(size)) {
-    switch (size) {
-      case 'big':
-        levers.size = 'su-px-34 su-py-15 su-text-20 md:su-text-24';
-        break;
+    levers.size = getButtonSize(size);
+  }
```
yvonnetangsu, line 50: Pulling button sizes into common levers/functions as well; I might do that with the variants also.

yvonnetangsu:
@sherakama I know I have one more task left to do for this one - allow custom icon. Let me know if you like this approach of sharing common levers and other code, and I'll finish up.

sherakama:
I'm having a read-through now. I won't block this PR for the added improvement of sending in an icon.

yvonnetangsu:
I think that for the Button component, instead of using the styles from the su-button class from the TW component, I'm just going to do everything in Decanter React with the TW utility classes - much easier dedupe and change, I think. I need to clean that up with the new Dismiss button changes. For the CTA Button, I didn't rely on the su-button class styles, and that works out better I think.

sherakama (approved):
GTG. 

I think what you have done here is fine. It has been broken up nicely and provides a good developer experience overall. The complexity and abstraction have been tucked away by helper functions that only the curious will need to dive down into. I think the documented options and inspecting the output will be enough for most developers. 

Nice work.

sherakama:
Also, I pushed a couple of changes as well. Please review those before merging.

yvonnetangsu:
> GTG.
> 
> I think what you have done here is fine. It has been broken up nicely and provides a good developer experience overall. The complexity and abstraction have been tucked away by helper functions that only the curious will need to dive down into. I think the documented options and inspecting the output will be enough for most developers.
> 
> Nice work.

Hey thank you for doing the revisions! Yeah, I've looked at the changes and have learned from your review of the Dismiss Button PR so will tackle props and classNames etc in a similar way in the future. Thanks for the new stories too 🚀

sherakama:
Yeah, I am trying to stick to convention with the props where we don't over complicate the simple components and anything that can go directly through to the HTML element as a prop does without us changing the key or messing with the values in between.

PR #32, opened 2021-05-07
ADAPT-1771: Convert string keys to ints.

sherakama:
# READY FOR REVIEW

# Summary
- Converts the keys in an object to integers and leaves the non-integer values as is. 

# Needed By (Date)

# Urgency
- How critical is this PR?

# Steps to Test

1. Do this
1. Then this
2. Then this

- JIRA ticket
- Other PRs

# See Also

### DIFF
--- a/src/Heading/Heading.js
+++ b/src/Heading/Heading.js
@@ -9,6 +9,7 @@ import {
   headingTracking,
   headingSizes,
 } from "./Heading.levers";
+import intKeys from "../common/util/intKeys";
 
 /**
  * Heading Component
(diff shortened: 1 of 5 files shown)


src/Heading/Heading.js
```diff
@@ -119,7 +120,7 @@ Heading.propTypes = {
    * Optional - to override the default size for the heading level selected.
    * Step 0 = base font size; each step up is a factor of 1.25 of the previous.
    */
-  size: PropTypes.oneOf(Object.keys(headingSizes)),
+  size: PropTypes.oneOf(Object.keys(headingSizes).map((val) => intKeys(val))),
```
yvonnetangsu, line 123: Does the heading level need this treatment too or is that one fine as is?

yvonnetangsu (approved):
I checked the preview site and everything still work the same as before. Approving - just that question about whether we do the same thing to the heading level also. Thanks!

src/Heading/Heading.js
sherakama: The level one doesn't because the levels are in an array and an array can have integer values. The problem with using the object keys is that they are always cast to strings.

yvonnetangsu: Ah, good to know. Thanks! GTG

PR #33, opened 2021-05-12
ADAPT-1776 Build Card Component

rebeccahongsf:
# READY FOR REVIEW

# Summary
- Build Card Component

# Needed By (Date)
- tbd

# Urgency
- low

# Steps to Test
1. Pull this PR
2. Run `npm run dev`
3. Review Card component and its variants

- [ADAPT-1776](https://stanfordits.atlassian.net/browse/ADAPT-1776)

# See Also


src/Card/Card.js
```diff
@@ -0,0 +1,159 @@
...
+        )}
+        <Heading
+          level={headingLevel ?? 3}
+          className={dcnb("su-font-serif su-bold su-mb-0")}
+        >
+          {headline}
+        </Heading>
+        {text && (
+          <p className="su-card-paragraph su-rs-mt-neg1 su-mb-0">{text}</p>
+        )}
```
yvonnetangsu, line 93: Looking great! I'll start adding comments (I realize you're still working on it, so please feel free to ignore).

I can see the card body here being something passed from a WYSIWYG for example, please take a look at the Alert body (I think it's the "children" prop in the Alert.js file) which has support for plain text as well as a WYSIWYG content object - could probably reuse that code 😃

src/Card/Card.js
```diff
@@ -0,0 +1,159 @@
...
+    bodyAlign = "su-items-center";
+  }
+
+  return (
+    <div
+      className={dcnb("basic-card su-max-w-600 su-basefont-23", wrapperClasses)}
+    >
+      {image && (
+        <div className={imageWrapperClasses} aria-hidden="true">
+          <img src={image.filename} alt={image.alt} />
```
yvonnetangsu, line 73: SODA actually recommends having `alt=""` for cards since 99% of the time, card images are decorative and have no direct relationship to the message of the card. For example, the image would be a smiling student with the card headline "Register for event X". I think we could safely take away the image.alt here and just set it to `alt=""`.

src/Card/Card.js
```diff
@@ -0,0 +1,159 @@
...
+        </div>
+      )}
+      <FlexBox
+        direction="col"
+        className={dcnb("card-body", bodyPadding, bodyAlign)}
+      >
+        {superHeadline && (
+          <span className={dcnb("su-font-serif su-bold su-mb-0")}>
```
yvonnetangsu, line 81: I imagine that it would be nice for the user to add custom classnames to each of these elements (superheadline, headline, button/link etc). Take a look at the Alert.js file and see how Shea created the "classes" object which allows for passing classnames into the individual elements 😄

src/Card/Card.js
```diff
@@ -0,0 +1,159 @@
...
+  // Levers
+  // ---------------------------------------------------------------------------
+
+  let wrapperClasses =
+    "su-bg-white su-text-black su-border su-border-solid su-border-transparent-black su-shadow";
+
+  let buttonClasses = "su-rs-mt-neg1";
+
+  // Basic card image has aspect ratio 4x3 for non-round option
+  let imageWrapperClasses = "su-aspect-w-4 su-aspect-h-3";
```
yvonnetangsu, line 47: Please check with @kerri-augenstein to see what she'd like the default aspect ratio to be, and whether we want options for different aspect ratios 😺

src/Card/Card.js
sherakama: I'd actually like to talk about that pattern a bit. The classes object works but isn't super elegant. It might make more sense to break down the components like how the headless UI library does it. Headless UI doesn't just have one import to make a component but rather, an outer most component and child components that you have to nest in your own template. see how they do radiogroup https://headlessui.dev/react/radio-group

Please explore this pattern for cards as I think it is a much more flexible pattern even if it requires more up-front work from the developer to 'get all the parts' right. Storbook and the documentation should help with that problem.

yvonnetangsu: Yeah that makes sense - would be much nicer to use the nested Heading and CTAs.

sherakama: Yeah, and you could swap them out with anything much easier. In this case a lot of our variants will be different assembly of the components and we will document them as 'stories' in storyblok.

src/Card/Card.stories.js
```diff
@@ -0,0 +1,74 @@
...
+      control: {
+        type: "boolean",
+      },
+    },
+  },
+};
+
+const CardTemplate = ({ ...rest }) => (
+  <Card {...rest}>
+    {/* Developer would pass in their own card content components */}
```
rebeccahongsf, line 27: Hmm. Not entirely sure if I interpreted what we had envisioned for the Card wrapper 🤔  But here's my updated/super simplified take on it — assuming we want to give developers more flexibility with passing ino other components. Although I'm not sure if this is the best way to display this example within Storybook.

Thoughts, feedback, suggestions @sherakama @yvonnetangsu

yvonnetangsu: I only took a quick look but yes I believe you're going in the right direction. Looking great 🎉 ! If you could please move it under the "Composite" group (currently under "Simple") that'd be awesome 😄

src/Card/Card.js
```diff
@@ -0,0 +1,85 @@
...
+  image,
+  align,
+  isMinimal,
+  ...props
+}) => {
+  // Defaults & Variables.
+  // ---------------------------------------------------------------------------
+
+  let wrapperClasses =
+    "su-bg-white su-text-black su-border su-border-solid su-border-transparent-black su-shadow";
```
yvonnetangsu, line 24: Quick tip: su-border-transparent-black is only available in the SAA site (we added this color in the the SAA homesite TW color theme which extends the Decanter v7 theme colors). You could possibly sub it using su-border-black-10 (which is 10% black in the Identity Guide) or even su-foggy-light - see which one Kerri signs off on 😜

src/Card/Card.stories.js
```diff
@@ -0,0 +1,75 @@
...
+    <span>Super headline</span>
+    <Heading>Headline</Heading>
+    <p>
+      Fusce convallis metus id felis luctus adipiscing. Vestibulum suscipit
+      nulla quis orci. Phasellus a est.
+    </p>
+    <CtaLink text="Call to action link" />
+  </Card>
```
yvonnetangsu, line 35: Please feel free to pass some props and attributes to these example children elements for a good "default" card, e.g., the headline level would most likely be an h3, and the size of the heading might be a 2 or 3. The body text could use the class `su-card-paragraph`. It would be good to also add `href` to the CTAs so the example will have a focus state and the right cursor 😃

yvonnetangsu: @rebeccahongsf Apologize I have a small correction to my comment - I think the size of the heading is more like 1 or 2 (I forget that 1 is already a step up from normal base font size (normal = 0, 1 = 1.25 x)) 😂

yvonnetangsu: And this is looking better and better - nice work 🥇 ! Let us know when you're ready for a real review so you have time to tweak and merge before sprint close 😃

rebeccahongsf:
Aside from needing to check with @kerri-augenstein in regards to the Card border shadow and image aspect ratio, I believe this PR is ready for review! 😄

src/Card/Card.js
```diff
@@ -0,0 +1,85 @@
...
+    >
+      {image && (
+        <div className={imageWrapperClasses} aria-hidden="true">
+          <img src={image.filename} alt="" />
+        </div>
+      )}
+      <FlexBox
+        direction="col"
+        className={dcnb("card-body", bodyPadding, bodyAlign)}
+      >
```
yvonnetangsu, line 59: For flexibility, we might want to allow users to add custom classes to the image wrapper and the card content wrapper also (eg, if they want to change the padding to the content). In that case, I wonder if we should break this up into more card subcomponents, e.g., `<CardContent>` and have user assemble it, or leave it like this, but tweak the custom classname support. Let's talk in adapt dev chat 😄

yvonnetangsu (changes requested):
I'll review some more later today but got a small list for you here. Thank you, @rebeccahongsf!

yvonnetangsu:
@rebeccahongsf Just remember one more nice to have item - please provide an example or two with the CtaButton instead of the CtaLink if you have time 🙏🏼

rebeccahongsf:
@kerri-augenstein has confirmed that

> (1) Image Aspect Ratio should be 2:1. See [Figma](https://www.figma.com/file/Kmd4utmJFPRMVeCFEEBQhLtx/Decanter-Design-System?node-id=60%3A2916) for additional info.
> (2) Let's go with su-border-black-10, so that when the card is on top of the light-fog background it has a slight border still. 

I'll include these style updates with my next commit! 😄

src/Card/Card.js
```diff
@@ -0,0 +1,93 @@
...
+        "basic-card su-max-w-600 su-basefont-23",
+        wrapperClasses,
+        className
+      )}
+    >
+      {image && (
+        <div className={imageWrapperClasses} aria-hidden="true">
+          <img src={image.filename} alt="" />
+        </div>
+      )}
```
sherakama, line 63: I'd like to see this as its own sub-component. As a developer, I would like to swap out the image component with a media component. Can you take a stab at a refactor where this becomes `<Card.Image>` instead?

src/Card/Card.stories.js
```diff
@@ -0,0 +1,98 @@
...
+    },
+  },
+};
+
+const CardTemplate = ({ ...rest }) => (
+  <Card {...rest}>
+    {/* Developer would pass in their own card content components */}
+    <span className="su-font-sans su-font-bold su-type-0 su-mb-0">
+      Super headline
+    </span>
```
sherakama, line 32: Thoughts about creating a `<Card.SuperHeadline>` sub-component for this?

src/Card/Card.stories.js
```diff
@@ -0,0 +1,98 @@
...
+    <Heading level="3" size={1} className="su-mb-0">
+      Headline
+    </Heading>
```
sherakama, line 35: It might be redundant now but if we wanted to add a context to the whole card later we might have to wrap this header in another component so that it can `useContext(CardContext)` so it might be best to wrap this in a `<Card.Heading>` sub-component.

src/Card/Card.stories.js
```diff
@@ -0,0 +1,98 @@
...
+    <CtaButton
+      text="Learn more"
+      srText="about Decanter"
+      animate="right"
+      icon="more"
+      variant="solid"
+      size="default"
+      href="https://decanter.stanford.edu"
+      className="su-mt-16"
+    />
```
sherakama, line 56: Would we want to wrap this in a containing HTML element so it is easier to move around later with css or no?

sherakama: I see a previous comment from Yvonne that agrees.

yvonnetangsu: I could see it being helpful to wrap it, so for example, if it's a minimal card, then no padding on this wrapper. Totally open to ideas.

yvonnetangsu: Haha we posted at the same time @sherakama

src/Card/Card.stories.js
```diff
@@ -0,0 +1,98 @@
+import React from "react";
+import { Card } from "./Card";
+import { cardAlignOptions } from "./Card.levers";
+import { Heading } from "../Heading/Heading";
+import { CtaLink } from "../CtaLink/CtaLink";
+import { CtaButton } from "../CtaButton/CtaButton";
+
+export default {
```
sherakama, line 8: Can I haz link to Figma too, please.

rebeccahongsf:
Ah ha!! I just had an epiphany for creating the subcomponents. Please disregard recent code changes while I refactor/rewrite. 😅

src/Card/Card.js
```diff
@@ -0,0 +1,158 @@
...
+    >
+      {children}
+    </article>
+  );
+};
+
+// Subcomponents
+// ---------------------------------------------------------------------------
+const Content = ({ children, className, image, align, isMinimal }) => {
```
rebeccahongsf, line 48: @sherakama @yvonnetangsu 🤠  I added these props from the `Card` wrapper so that the example(s) would work within Storybook but I wonder if there's a more elegant way to pass props from the `<Card>` to `<Card.Content>` or if props should be passed in this way at all.

sherakama: Props shouldn't be passed down into the sub-components here. Each component and sub-component should have its own set of props. If multiple components need to "talk" to each other or know about each other's state, you will need to create a context. 


HeadlessUI is TS but you can see that Group, Switch, Label, and Description all have their own props. 
See: https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-react/src/components/switch/switch.tsx#L37

rebeccahongsf:
It appears that the subcomponents built within the Card components does not have their argTypes displayed within the Storybook documentation

image

I believe it's because the subcomponents' argsTypes aren't defined within their own respective `stories.js` 🤔

src/Card/Card.js
```diff
@@ -0,0 +1,206 @@
...
+    PropTypes.object,
+  ]),
+
+  // Children
+  children: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.element,
+    PropTypes.node,
+  ]),
+};
```
rebeccahongsf, line 196: Should subcomponents have their PropTypes defined? 🤔

sherakama:
> Should subcomponents have their PropTypes defined? 🤔

Yes

src/Card/Card.js
```diff
@@ -0,0 +1,206 @@
...
+    "su-bg-white su-text-black su-border su-border-solid su-border-black-10 su-shadow";
+
+  if (isMinimal) {
+    wrapperClasses = "";
+  }
+
+  return (
+    <article
+      className={dcnb(
+        "basic-card su-max-w-600 su-basefont-23",
```
sherakama, line 28:
```suggestion
        "card su-max-w-600 su-basefont-23",
```
To match the component.

src/Card/Card.levers.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * Styles for the card.
+ */
+// export const cardVariants = ["solid", "outline", "ghost", "unset"];
+
+export const cardAlignOptions = ["left", "center"];
```
sherakama, line 6: How much work is it to add the right aligned option as well? It just seems like it is missing.

src/Card/Card.js
```diff
@@ -0,0 +1,206 @@
...
+
+  let wrapperClasses =
+    "su-bg-white su-text-black su-border su-border-solid su-border-black-10 su-shadow";
+
+  if (isMinimal) {
+    wrapperClasses = "";
+  }
+
+  return (
+    <article
```
sherakama, line 26: It probably makes sense to make the outer wrapper element configurable here. eg: [div, span, li, article, etc]

src/Card/Card.js
```diff
@@ -0,0 +1,206 @@
...
+    >
+      {children}
+    </FlexBox>
+  );
+};
+
+Content.displayName = Content;
+Card.Content = Content;
+
+const Image = ({ image, className }) => {
```
sherakama, line 64: It doesn't make sense to make image an object here if you are only using one property. Two paths forward for you. 

1. Continue to use image as an object and destructure what you want out of it for `src` but then expand the rest of the props on the `img` element so users can modify the values. 
2. Change the image prop to a string type and rename `filename`

sherakama: You destructure the className prop but don't use it anywhere.

src/Card/Card.js
```diff
@@ -0,0 +1,206 @@
...
+  return (
+    <div className={imageWrapperClasses} aria-hidden="true">
+      <img src={image.filename} alt="" />
+    </div>
+  );
+};
+Image.displayName = Image;
+Card.Image = Image;
+
+const Superhead = ({ children }) => (
```
sherakama, line 77:
```suggestion
const Superhead = ({ children, className }) => (
```

src/Card/Card.js
```diff
@@ -0,0 +1,206 @@
...
+  <span className="su-font-sans su-font-bold su-type-0 su-mb-0">
```
sherakama, line 78:
```suggestion
  <span className={dcnb("su-font-sans su-font-bold su-type-0 su-mb-0", className)}>
```

src/Card/Card.js
```diff
@@ -0,0 +1,206 @@
...
+Superhead.displayName = Superhead;
+Card.Superhead = Superhead;
+
+const Headline = ({ children }) => <Heading>{children}</Heading>;
+Headline.displayName = Headline;
+Card.Headline = Headline;
+
+const Body = ({ children }) => <p className="su-card-paragraph">{children}</p>;
+Body.displayName = Body;
+Card.Body = Body;
```
sherakama, line 91: This body seems too restrictive. What if I wanted multiple paragraphs or a list or something? I would suggest removing it altogether and let the developer pass in their content as they see fit directly in the Card components children.

src/Card/Card.js
```diff
@@ -0,0 +1,206 @@
...
+
+const Cta = ({ children, className }) => (
+  <div className={className}>{children}</div>
+);
+Cta.displayName = Cta;
+Card.Cta = Cta;
```
sherakama, line 97: This seems a bit redundant as a developer I could bypass this component and just put this content directly in the Card component's children. Unless you want to provide more structure I would suggest on removing it.

src/Card/Card.levers.js
rebeccahongsf: Not much 😄  The option has been added!

sherakama: ty

src/Card/Card.stories.js
```diff
@@ -0,0 +1,355 @@
...
+MinimalImage.storyName = "Minimal with Image";
+
+export const MinimalCenter = CardTemplateMinimalCenter.bind({});
+MinimalCenter.args = {
+  isMinimal: true,
+};
+MinimalCenter.storyName = "Minimal Center Content";
+
+export const MinimalRight = CardTemplateMinimalRight.bind({});
+MinimalCenter.args = {
```
sherakama, line 352:
```suggestion
MinimalRight.args = {
```

src/Card/Card.js
```diff
@@ -0,0 +1,190 @@
...
+  <span
+    className={dcnb("su-font-sans su-font-bold su-type-0 su-mb-0", className)}
+  >
+    {children}
+  </span>
+);
+Superhead.displayName = Superhead;
+Card.Superhead = Superhead;
+
+const Headline = ({ children }) => <Heading>{children}</Heading>;
```
sherakama, line 98:
```suggestion
const Headline = ({ children, level, size, className }) => <Heading level={level ?? "3"} size={size ?? 1} className={dcnb("su-mb-0", className)}>{children}</Heading>;
```

sherakama: This might not be correct syntax as I am freestyling here.

sherakama: Ignore this. Doing it in props.

sherakama: Here is the work now. 
https://github.com/SU-SWS/decanter-react/pull/48

sherakama:
Ok, fixed the linting. Green lights all around. Marking as GTG. @yvonnetangsu for final pass and pressing of the green button.

sherakama (approved):
Awesome work all.

src/Card/Card.Superhead.js
```diff
@@ -0,0 +1,71 @@
...
+ * Superhead component
+ * @param {*} param0
+ * @returns
+ */
+
+export const Superhead = ({
+  weight,
+  className,
+  children,
+  isUppercase,
```
rebeccahongsf, line 16: @yvonnetangsu isUppercase boolean has been added to the Superhead subcomponent ٩ʕ◕౪◕ʔو

yvonnetangsu (approved):
Yay - GTG! Thanks for the uppercase prop. Party time 🥳 !

PR #44, opened 2021-05-19
ADAPT-2517: Follow-up tweaks for auto-release action.

nplowman:
# READY FOR REVIEW

# Summary
- Removes 'v' prefix from release tags.
- Adds release notes from changelog to Github Release
- Automatically flags alpha, beta, and rc releases as "Prerelease" releases in Github.
- Other releases are published as "latest release".

# Needed By (Date)
- Next 1-3 days.

# Urgency
- 5/10 - Useful process improvement, but not urgent.

# Steps to Test
As with previous work on this ticket, a PR must be merged with a label in order to trigger the workflow. 
So after merging this ticket, you may want to make a few PRs with edits to the README file or some other innocuous change to test the different events.

1. Merge a PR with the 'alpha', 'beta' or 'rc' label. Validate that release in github is flagged as "Prerelease"
1. Merge a PR with the 'patch', 'minor', or 'major' label. Validate that github release is marked as "Latest Release"
1. For any PR with a release label, validate that Github release contains release notes. 

- https://stanfordits.atlassian.net/browse/ADAPT-2517

# See Also


.github/workflows/autorelease.yml
```diff
@@ -89,7 +95,33 @@ jobs:
...
+      - name: Set enviroment variables for release notes.
+        run: |
+          echo "RELEASE_TAG=$(git tag --sort=-taggerdate | head -1)" >> $GITHUB_ENV
+          echo "PREVIOUS_TAG=$(git tag --sort=-taggerdate | sed -n '2 p')" >> $GITHUB_ENV
+
+      - name: Create Release Notes
+        id: release_notes
+        run: |
+          # Scrape release notes for this release from the full changelog.
+          release_notes=$(./node_modules/.bin/auto-changelog --stdout | awk "/\[$RELEASE_TAG\]/{flag=1; next} /\[$PREVIOUS_TAG\]/{flag=0} flag")
```
sherakama, line 108: `auto-changelog` has a `--unreleased-only` flag that should allow you to generate the release notes without having to scrape them.

nplowman: I'll give this a try and see if it works... 
I was trying with other flags, like --starting-version, but those weren't providing the expected output.
Also tried 2-3 different Github actions for auto-generating release notes and struck out with those as well.

nplowman: @sherakama Looks like --unreleased-only gets the correct messages if I move it _before_ the npm version command:
Screen Shot 2021-05-19 at 3 01 45 PM

However, we'd still need to do some post-processing to remove the additional lines at the beginning or pass in a custom template. 

Would you prefer one of those options, or should we stick with the existing solution?

sherakama: Thanks for looking into this. I don't want this to be more make work for you. I'll noodle on this a little more and if I don't come up with a compelling reason to change I think what you have is great.

PR #46, opened 2021-06-01
ADAPT-000: React 17 Support

sherakama:
# READY FOR REVIEW

# Summary
- Adds support for React V17 and keeps 16.

# Needed By (Date)
- Whenever?

# Urgency
- Low

# Steps to Test

1. Review build still builds
2. Profit


yvonnetangsu (approved):
Profit! 😄  The way you do it, looks like it would be ok if a project with React 16 is pulling components from here right?

sherakama:
> Profit! 😄 The way you do it, looks like it would be ok if a project with React 16 is pulling components from here right?

Yep, This package should support react `16.14.0` through `<17.1.0` with this syntax. I could probably simplify this to `>= 16.14.0` but it gives us some constraints for now.

PR #47, opened 2021-06-01
ADAPT-1776  Card possible tweaks

yvonnetangsu:
# READY
- (Edit the above to reflect status)

# Summary
- TL;DR - what's this PR for?

# Needed By (Date)

# Urgency
- How critical is this PR?

# Steps to Test

1. Do this
1. Then this
2. Then this

- JIRA ticket
- Other PRs

# See Also


src/Card/Card.stories.js
```diff
@@ -288,9 +274,7 @@ const CardTemplateMinimalRight = ({ ...rest }) => (
   <Card {...rest}>
     <Card.Image image={{ filename: "https://placekitten.com/600/300" }} />
     <Card.Content align="right" className="su-rs-pt-2">
-      <Card.Superhead className="su-font-sans su-font-bold su-type-0 su-mb-0">
-        Super headline
-      </Card.Superhead>
+      <Card.Superhead weight="regular">Super headline</Card.Superhead>
```
yvonnetangsu, line 277: @rebeccahongsf you could safely take the classnames/props out if they're the same as default, and just add the props/classnames that are different for this example 😃

src/Card/Card.js
```diff
@@ -85,17 +86,12 @@ const Image = ({ image, className }) => {
...
-  >
-    {children}
-  </span>
-);
 Superhead.displayName = Superhead;
 Card.Superhead = Superhead;
 
-const Headline = ({ children }) => <Heading>{children}</Heading>;
+const Headline = ({ children, ...props }) => (
+  <Heading {...props}>{children}</Heading>
```
yvonnetangsu, line 93: Added ...props into the Headling because we need them 😸  Now the level/size props should work as expected

src/Card/Card.Superhead.js
```diff
@@ -0,0 +1,42 @@
...
+      {children}
+    </span>
+  );
+};
+
+Superhead.propTypes = {
+  /**
+   * Font weight.
+   */
+  weight: PropTypes.oneOf(Object.keys(superheadWeights)),
```
yvonnetangsu, line 23: FYI adding comments in propTypes using this format will add help text to the Storybook interface, but using // won't.

src/Card/Card.Content.js
```diff
@@ -0,0 +1,46 @@
...
+export const Content = ({ children, className, align, ...props }) => {
+  // Content alignment including image and CTA, default is left-align
+  // This setting overrides the alignment option in the nested CTA
+  let bodyAlign = "su-items-start";
+  let contentClass = "";
+
+  if (align === "center") {
+    contentClass = dcnb(contentClass, "children:su-mx-auto su-text-center");
+    bodyAlign = "su-items-center";
+  }
```
yvonnetangsu, line 15: @rebeccahongsf  I just realized 2 things. The plugin that enables the "children" prefix isn't installed in Decanter 7 core. I'm going to open a PR to add that. Also, what do you think about moving the align props to the main "Card" component, since there might be cases that we need to include the alignment of the image (if it's a small image that doesn't span the width of the whole card) 😄

sherakama:
Hmm. Looks like you are separating the sub components out. Not that I have a problem with that but wondering why now and not before?

I have a conflicting PR with some similar work for this as well. 
https://github.com/SU-SWS/decanter-react/pull/48

yvonnetangsu:
> Hmm. Looks like you are separating the sub components out. Not that I have a problem with that but wondering why now and not before?
> 
> I have a conflicting PR with some similar work for this as well.
> #48

I haven't had a lot of time to spend on this until today. I was testing and looks like separating these Card.Content type components out solve the issue with the props not displaying for subcomponents (they're now showing)
Screen Shot 2021-06-01 at 3 07 23 PM

Also, I feel like this makes the main `Card.js` file smaller and easier to read since there's a lot going on. I'd rather have the code for the subcomponents in their own files to keep it more manageable 🤔

yvonnetangsu:
@sherakama I've merged your Card tweaks into my branch and our changes are now combined. Loading nicely on my local. I'll add a bit more later tonight to the Card.Image.

src/Card/Card.levers.js
```diff
@@ -7,4 +7,23 @@ export const cardAlignOptions = ["left", "center", "right"];
 /**
  * Available HTML elements of the container.
  */
-export const cardElements = ["div", "span", "article", "li"];
```
yvonnetangsu, line 10: @rebeccahongsf I'm removing "span" and "li" since "span" is a bit more suitable for inline/1-liner text, and "li" could be a bit dangerous to use on its own as the outer wrapper. I appreciate your consideration for accessibility (which is probably why you added the "li" option), but it's a bit safer if they could wrap the whole card within an `<li>` instead. Also, these 2 elements can introduce some unwanted base styles into the component 😄

sherakama: I am ok with `span` going away but Caryl just had a developer session on having cards in a list as `li` elements. I think we should maintain that one.

src/Card/Card.Content.js
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+export const Content = ({ children, className, align, ...props }) => {
+  // Content alignment including image and CTA, default is left-align
+  // This setting overrides the alignment option in the nested CTA
+  let bodyAlign = "su-items-start";
+  let contentClass = "";
+
+  if (align === "center") {
+    contentClass = dcnb(contentClass, "su-text-center");
```
sherakama, line 20:
```suggestion
    contentClass = "su-text-center";
```

contentClass at this point cannot be anything other than an empty string. No sense in running `dcnb` on it.

src/Card/Card.Content.js
```diff
@@ -0,0 +1,62 @@
...
+    bodyAlign = "su-items-center";
+  }
+
+  if (align === "right") {
+    contentClass = dcnb(contentClass, "su-text-right");
```
sherakama, line 25:
```suggestion
    contentClass = "su-text-right";
```

src/Card/Card.Headline.js
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Headline component
+ * @param {*} param0
+ * @returns
+ */
+
+export const Headline = ({
+  children,
+  level,
+  size,
```
sherakama, line 22:
```suggestion
  level = 3,
  size = 1,
```

src/Card/Card.Headline.js
```diff
@@ -0,0 +1,79 @@
...
+  weight,
+  font,
```
sherakama, line 24: Probably want to set defaults for `weight` and `font` as well unless you don't expect the developer to always use these.

src/Card/Card.Content.js
```diff
@@ -0,0 +1,62 @@
...
+    <FlexBox
+      direction="col"
+      className={dcnb("card-body", bodyAlign, contentClass, className)}
+      {...props}
+    >
+      {children}
+    </FlexBox>
+  );
+};
+
```
sherakama, line 39:
```suggestion

Content.displayName = "Card.Content";
```

src/Card/Card.Headline.js
sherakama: never mind. I see the default props.

src/Card/Card.Headline.js
sherakama: Nevermind I see the default props.

src/Card/Card.Headline.js
```diff
@@ -0,0 +1,79 @@
...
+    size={size}
+    font={font}
+    weight={weight}
+    {...props}
+    className={dcnb("su-mb-03em", className)}
+  >
+    {children}
+  </Heading>
+);
+
```
sherakama, line 39:
```suggestion

Headline.displayName = "Card.Headline";

```

src/Card/Card.Image.js
```diff
@@ -0,0 +1,50 @@
...
+
+  // Remove alt.
+  if (image.alt) {
+    // eslint-disable-next-line no-param-reassign
+    delete image.alt;
+  }
+
+  return (
+    <div className={dcnb(imageAspectRatio, className)} aria-hidden="true">
+      <img {...image} alt="" className="su-object-cover" />
```
sherakama, line 23: If you're going to add a classname to this element you will have to remove `className` from the `image` prop to avoid duplication like we did with `alt`.

src/Card/Card.Superhead.js
```diff
@@ -0,0 +1,52 @@
...
+        superheadWeight,
+        className
+      )}
+      {...props}
+    >
+      {children}
+    </span>
+  );
+};
+
```
sherakama, line 28:
```suggestion

Superhead.displayName = "Card.Superhead";

```

src/Card/Card.Superhead.js
sherakama: Most interesting. Thanks for sharing.

src/Card/Card.js
```diff
@@ -31,146 +33,59 @@ export const Card = ({ className, children, element, isMinimal, ...props }) => {
   return (
     <Element
       className={dcnb(
-        "card su-max-w-600 su-basefont-23",
+        "basic-card su-block su-max-w-600 su-basefont-23",
```
sherakama, line 36:
```suggestion
        "card su-block su-max-w-600 su-basefont-23",
```

src/Card/Card.js
```diff
@@ -31,146 +33,59 @@ export const Card = ({ className, children, element, isMinimal, ...props }) => {
...
-  if (align === "center") {
-    contentClass = dcnb(contentClass, "children:su-mx-auto su-text-center");
-    bodyAlign = "su-items-center";
-  }
 
-  if (align === "right") {
-    contentClass = dcnb(contentClass, "children:su-mx-auto su-text-right");
-    bodyAlign = "su-items-end";
-  }
+Image.displayName = "Card.Image";
```
sherakama, line 52:
```suggestion
```

src/Card/Card.js
```diff
@@ -31,146 +33,59 @@ export const Card = ({ className, children, element, isMinimal, ...props }) => {
...
-  return (
-    <FlexBox
-      direction="col"
-      className={dcnb("card-body", bodyAlign, contentClass, className)}
-    >
-      {children}
-    </FlexBox>
-  );
-};
+Content.displayName = "Card.Content";
```
sherakama, line 54:
```suggestion
```

src/Card/Card.js
```diff
@@ -31,146 +33,59 @@ export const Card = ({ className, children, element, isMinimal, ...props }) => {
...
 
-Content.displayName = Content;
-Card.Content = Content;
+Superhead.displayName = "Card.Superhead";
```
sherakama, line 56:
```suggestion
```

src/Card/Card.js
```diff
@@ -31,146 +33,59 @@ export const Card = ({ className, children, element, isMinimal, ...props }) => {
...
 
-const Image = ({ image, className }) => {
-  // Basic card image has aspect ratio 2x1
-  const imageWrapperClasses = "su-aspect-w-2 su-aspect-h-1";
+Headline.displayName = "Card.Headline";
```
sherakama, line 58:
```suggestion
```

src/Card/Card.levers.js
```diff
@@ -7,4 +7,23 @@ export const cardAlignOptions = ["left", "center", "right"];
...
+export const cardElements = ["div", "article"];
+
+/**
+ * Available font weights.
+ */
+export const superheadWeights = {
```
sherakama, line 15: Nice.

src/Card/Card.levers.js
```diff
@@ -7,4 +7,23 @@ export const cardAlignOptions = ["left", "center", "right"];
...
+  regular: "su-font-regular",
+  semibold: "su-font-semibold",
+  bold: "su-font-bold",
+};
+
+/**
+ * Available image aspect ratios.
+ */
+export const imageAspectRatios = {
```
sherakama, line 24: Good.

src/Card/Card.stories.js
```diff
@@ -28,39 +28,24 @@ export default {
   },
   argTypes: {
     element: {
-      control: {
```
sherakama, line 31: From what console says, control.options is deprecated and will be removed in version `7`.

src/Card/Card.stories.js
```diff
@@ -318,6 +301,45 @@ const CardTemplateMinimalRight = ({ ...rest }) => (
     </Card.Content>
   </Card>
 );
+
+const CardTemplateCustom = ({ ...rest }) => (
```
sherakama, line 305:
```suggestion
const CardTemplateCustom = (props) => (
```

src/Card/Card.stories.js
```diff
@@ -318,6 +301,45 @@ const CardTemplateMinimalRight = ({ ...rest }) => (
...
+  <Card {...rest} className="su-bg-foggy-light" data-label="card">
```
sherakama, line 306:
```suggestion
  <Card {...rest} className={dcnb("su-bg-foggy-light", className)} data-label={dataLabel || "card"}>
```

sherakama, line 306: should the attribute `data-label` go directly on the card component and not the story. Also, why add this attribute?

sherakama: Oh, this is just for the custom option. Nevermind.

src/Card/Card.levers.js
yvonnetangsu: Huh...using `li` as the direct outer wrapper? Hmmm....ok. Let's keep it then. I think it does introduce some style changes. Let me double check.

src/Card/Card.Content.js
yvonnetangsu: I actually am thinking about moving `align` to the parent Card.Root, since sometimes alignment needs to include the image (eg, if they swap out the full width card image with a small thumbnail like we did on SAA.

src/Card/Card.stories.js
yvonnetangsu: This one is just for the example template

src/Card/Card.stories.js
yvonnetangsu: Good to know...

src/Card/Card.levers.js
yvonnetangsu: Added `li` back, and line height class so there will be no style change when using `li`

src/Card/Card.Image.js
yvonnetangsu: Done

src/Card/Card.Content.js
yvonnetangsu: Hmmm.....nevermind. Until we add the children variant plugin to the Decanter 7 repo, it's simpler to leave the align prop in the Card.Content subcomponent. Maybe later 🤔

yvonnetangsu:
@rebeccahongsf This branch has the combined tweaks that @sherakama and I suggested. Please pull down locally to take a look at the examples and the doc view (the React code are now looking good)  
Note: you might have to do a `npm run build` after pulling this branch, since I added some new TW classes so it needed another CSS purge. Quick way to check, if you see this custom card showing a square image, that means it has built properly (it has the `su-aspect-w-1` class which previously wasn't there).
Screen Shot 2021-06-01 at 11 31 17 PM

If you like what you see, then please feel free to merge this branch into your card branch 😃 . For the examples, I kept either the CTA link or CTA button in each card, as that's closer to the real thing (don't want to give people ideas that they could add multiple CTAs 😂 ).

if you have time in the morning and looking for stuff to do, please feel free to add a boolean prop in the superhead subcomponent for uppercase, as we've used uppercase superheadline quite often. If you're busy with some other things, that can always wait - totally just a nice to have.

Thank you for your nice work!! 👏🏼

src/Card/Card.Content.js
rebeccahongsf: Sounds good! I didn't think of the use case of a small card that doesn't span the width of the whole card. 🤔  We can definitely bump the alignment props up. My question though would be did we want the image and body content to inherit the alignment from the main Card component or did we want to have each subcomponent aligned separately (i.e. the developer can then chose to align the image right but have the body centered/left aligned)?

rebeccahongsf:
> @rebeccahongsf This branch has the combined tweaks that @sherakama and I suggested. Please pull down locally to take a look at the examples and the doc view (the React code are now looking good)
> Note: you might have to do a `npm run build` after pulling this branch, since I added some new TW classes so it needed another CSS purge. Quick way to check, if you see this custom card showing a square image, that means it has built properly (it has the `su-aspect-w-1` class which previously wasn't there).
> Screen Shot 2021-06-01 at 11 31 17 PM
> 
> If you like what you see, then please feel free to merge this branch into your card branch 😃 . For the examples, I kept either the CTA link or CTA button in each card, as that's closer to the real thing (don't want to give people ideas that they could add multiple CTAs 😂 ).
> 
> if you have time in the morning and looking for stuff to do, please feel free to add a boolean prop in the superhead subcomponent for uppercase, as we've used uppercase superheadline quite often. If you're busy with some other things, that can always wait - totally just a nice to have.
> 
> Thank you for your nice work!! 👏🏼

Everything looks absolutely amazing! Thank you for fixing and refactoring the subcomponents and adding an additional custom example @yvonnetangsu and @sherakama !!! I shall merge this into my PR and work real quick on adding the uppercase boolean for the Superhead subcomponent. 🥳

src/Card/Card.Content.js
yvonnetangsu: @rebeccahongsf Don't worry about this for now 😄 ! I found that this is a lot easier if we have the "children" prefix plugin like we do in the SAA repo, and you got a good point about perhaps people might want the image and content to align differently. Let's table this for now. Your current way stays true to the spirit of the separate subcomponents 😸

PR #50, opened 2021-06-08
ADAPT-2695 Create Poster component

rebeccahongsf:
# READY FOR REVIEW

# Summary
- Create Poster component

# Needed By (Date)
- tbd

# Urgency
- low

# Steps to Test

1. Pull this PR
2. Run `npm run dev` or view [component on netlify here](https://deploy-preview-50--decanter-react.netlify.app/?path=/story/composite-poster--default)
3. Review code and component display 

- [ADAPT-2695](https://stanfordits.atlassian.net/browse/ADAPT-2695)

# See Also


src/Poster/Poster.js
```diff
@@ -0,0 +1,96 @@
...
+  if (layout === "left") {
+    wrapperClasses = "su-flex su-flex-col su-justify-center md:su-flex-row";
+    imageWrapper =
+      "first:su-min-w-[14rem] su-rs-mb-2 su-mx-auto md:su-rs-mr-4 md:su-mb-0 md:su-ml-0";
+    contentWrapper = "last:su-items-start md:su-flex-grow";
+  } else {
+    wrapperClasses = "su-flex su-flex-col su-items-center";
+    contentWrapper = "last:su-items-center su-text-center";
+    imageWrapper = "first:su-rs-mb-2";
+  }
```
yvonnetangsu, line 35: I think you're totally on the right track. If you worry about spacing above/below the different parts, it's ok to put those in the subcomponents 😄  (<- not sure if that's what you were concerned about). 

FYI (and you probably already know this), the first/last prefixes and the TW classes with [] are not available for this repo sadly 😢 

Let's check with everyone if we want to include background image in this component for now - don't want to scope creep on ya 😂  Please feel free to post to the adapt channel - let's have a discussion on what background color/text colors to include so you don't end up doing more work than needed.

src/Poster/Poster.stories.js
```diff
@@ -0,0 +1,71 @@
...
+      options: layoutOptions,
+    },
+  },
+};
+
+const PosterTemplate = ({ ...rest }) => (
+  <Poster {...rest}>
+    <Poster.Content>
+      <Poster.Headline className="su-mb-0">Title</Poster.Headline>
+      <p className="su-big-paragraph children:su-leading-snug">
```
yvonnetangsu, line 37: children prefix also not available 😢

src/common/util/addBgImage.js
```diff
@@ -0,0 +1,33 @@
...
+  if (imageUrl != null) {
+    if (imageUrl.startsWith("http")) {
+      const originalWidth = getImageWidth(imageUrl);
+
+      // Downsize image if it's wider than 2000px, otherwise just reduce jpg quality to 60%
+      if (originalWidth > 2000) {
+        imgSrc = imageUrl.concat("/2000x0");
+      } else {
+        imgSrc = imageUrl;
+      }
```
yvonnetangsu, line 21: For this repo, totally ok to take out image processing (resize, change jpg quality etc) since these were added to use with Storyblok's image service. For this repo, I think we can leave it up to the user to pass in the image processed on their end.

yvonnetangsu:
@rebeccahongsf Looking better and better. A few ideas - please feel free to push back:

- Let's use the foggy light poster as the first example story, since the white poster blends in with the page background. Also, I suspect a lot of websites have white backgrounds, so having a light grey poster would make it actually look like a poster instead of just a WYSIWYG among the rest of the page content 😄 
- Perhaps some examples with the various CTA buttons in the poster body? (white poster with solid red, black with the ghost button sound pretty good 👻 )
- A couple headline and body copy text size options like we have in SAA site would be nice, in examples totally fine.
- This is from Kerri from previous SAA communications, for `su-card-paragraph` sized body copy, she'd like the line height to be 1.3 (su-leading-snug).
- Let's kill the bottom margin on the last element in the poster.

Thanks in advance!

src/Poster/Poster.js
```diff
@@ -0,0 +1,115 @@
...
+
+  let gradientOverlay = "linear-gradient(to bottom, transparent, #181D1C)";
```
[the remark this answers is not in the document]
rebeccahongsf: I had set a default `linear-gradient` for the `gradientOverlay` variable in the event that the developer forgets to pass in a gradient.

Does it make sense to have a gradient option (lever)? 
Alternatively, we could remove the default gradient and have it as an empty string instead. That way it would be completely in the developer's hands to decide whether they want an overlay on their `bgImage`. Thoughts? 😄

src/Poster/Poster.js
rebeccahongsf: Realized that `first:` and `last:` prefixes do not work within out decanter-react repo for flex styles (margin only but even then it's a bit finicky). 😢  I'll rewrite to include the alignments within the stories and remove the lever. Perhaps we can revisit adding the alignment lever if/when we expand the coverage of `first:` and `last:` prefixes?

rebeccahongsf: Or we can keep the layout lever since it's could be seen as image left or center! The body content will be left align unless additional classes are passed to the Content subcomponent.

src/Poster/Poster.js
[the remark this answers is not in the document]
sherakama: I was more thinking of an on/off lever for bg gradients vs options. What if the developer wants no gradient? Can you make it easy to remove?

sherakama (approved):
This is great. Thanks for the work. I refactored the Poster. Image component a bit as the image.src object requirement was too restrictive for using things like Gatsby's image optimization components. I left the background image alone as that is less likely to need a component to optimize the image.

Screen Shot 2021-06-09 at 11 08 28 PM


Screen Shot 2021-06-09 at 11 08 50 PM

src/Poster/Poster.js
```diff
@@ -0,0 +1,115 @@
...
+  layout,
+}) => {
+  let wrapperClasses;
+  let imageWrapper;
+  let contentWrapper;
+
+  if (layout === "left") {
+    wrapperClasses = "su-flex su-flex-col su-justify-center md:su-flex-row";
+    imageWrapper =
+      "first:su-min-w-[14rem] su-rs-mb-2 su-mx-auto md:su-rs-mr-4 md:su-mb-0 md:su-ml-0";
```
sherakama, line 30:
```suggestion
      "su-rs-mb-2 su-mx-auto md:su-rs-mr-4 md:su-mb-0 md:su-ml-0";
```

Sorry, no JIT compiler here.

src/Poster/Poster.js
```diff
@@ -0,0 +1,115 @@
...
+   *  Is the layout alignment center or left?
+   *
+   */
+  layout: PropTypes.oneOf(layoutOptions),
+
+  // Background Image
+  bgImage: PropTypes.shape({
+    src: PropTypes.string,
+  }),
+};
```
sherakama, line 109: Missing props for bgGradient and bgColor

src/Poster/Poster.stories.js
```diff
@@ -0,0 +1,230 @@
...
+    <Poster.Image className="su-rs-mb-2">
+      <img
+        src="https://placekitten.com/600/400"
+        className="su-object-cover su-w-full su-h-full su-object-center"
+        alt=""
+      />
+    </Poster.Image>
+    <Poster.Content className="su-text-white su-items-center su-text-center su-align-center">
+      <Poster.Headline>Title</Poster.Headline>
+      <p className="su-big-paragraph su-leading-snug">
```
yvonnetangsu, line 101: If you could please add a `su-mb-0` to last paragraphs in all the examples, that'd be great.

src/Poster/Poster.stories.js
```diff
@@ -0,0 +1,230 @@
...
+        href="https://decanter.stanford.edu"
+        className="su-rs-mt-0 su-bg-white hocus:su-bg-lagunita su-text-black hocus:su-text-white su-px-20 su-pt-14 su-pb-16"
+      />
+    </Poster.Content>
+  </Poster>
+);
+
+const PosterTemplateCustomBorder = ({ ...rest }) => (
+  <Poster {...rest}>
+    <Poster.Image className="su-border-3 su-border-solid su-border-digital-red-light su-rs-mb-2">
```
yvonnetangsu, line 144: Looks like Figma image border is a 7px one, but Decanter v7 doesn't have a 7px width currently. I know @sherakama added some recently, so let's go with either 6px or 8px 😄  Although... @sherakama , is there a release that got punched with your border width PR?
https://github.com/SU-SWS/decanter/pull/834/files

yvonnetangsu (changes requested):
Just those few final minor requests, and one more here:

For the "default" (the first example), let's add some classes so the content is center-aligned (it's a bit more commonly used), and let's add a solid red CTA button with it 😄 

After that, then your work is done - great job 👏🏼  And thank you for your help in refactoring too, @sherakama !

yvonnetangsu:
@rebeccahongsf One minor note - if you could use the "large" (or was it "big") CTA button `size` prop for the poster examples, that'd be awesome 🙏🏼

src/Poster/Poster.stories.js
rebeccahongsf: The latest release does not seem to include the border-width 😢  Currently the available options are
```
.su-border-0{
  border-width: 0
}

.su-border-2{
  border-width: 2px
}

.su-border-3{
  border-width: 3px
}

.su-border{
  border-width: 1px
}
```

sherakama: Send in a PR to add border-7 and add the `alpha` label to it. That should get you what you need.

sherakama:
@yvonnetangsu I think this is GTG now too. Anything left?

yvonnetangsu:
> @yvonnetangsu I think this is GTG now too. Anything left?

Very very close - if we could center align the "default" example and add a CTA button, and use the big CTA button for the last teal poster example then we're GTG 😄

rebeccahongsf:
@yvonnetangsu examples have been updated! 😄  Let me know if any other adjustments or examples are needed

yvonnetangsu:
> @yvonnetangsu examples have been updated! 😄 Let me know if any other adjustments or examples are needed

Thank you~ If you could please use the large/big button size for the black and lagunita posters, then we're GTG. We can deal with the border width on the image later - I think Shea suggested open a PR in Decanter v7 to add border width 7px, but that can totally wait 😸

src/Poster/Poster.stories.js
```diff
@@ -0,0 +1,249 @@
...
+        faucibus, nunc et pellentesque egestas, lacus ante convallis tellus,
+        vitae iaculis lacus elit id tortor.
+      </p>
+      <CtaButton
+        text="Learn more"
+        srText="about Decanter"
+        animate="right"
+        icon="more"
+        variant="unset"
+        size="small"
```
yvonnetangsu, line 153:
```suggestion
        size="big"
```

yvonnetangsu (approved):
GTG - thank you for the changes! Nice work 🎉 😸  We can add that image border width change later after the Decanter PR get merged. Thank you!

yvonnetangsu:
@sherakama Got this email - "Auto Release: All jobs have failed", but it says all 11 checks passed. Any ideas? 🤔

sherakama:
Looks like the release failed to build https://github.com/SU-SWS/decanter-react/runs/2824551382?check_suite_focus=true 

Could be missing export or missing `npm run package`

PR #51, opened 2021-06-08
ADAPT-2565

romakravchuk:
# READY FOR REVIEW/NOT READY
- (Edit the above to reflect status)

# Summary
- TL;DR - what's this PR for?

# Needed By (Date)

# Urgency
- How critical is this PR?

# Steps to Test

1. Do this
1. Then this
2. Then this

- JIRA ticket
- Other PRs

# See Also

### DIFF
--- a/src/Alert/Alert.js
+++ b/src/Alert/Alert.js
@@ -4,6 +4,7 @@ import Icon from "react-hero-icon";
 import { dcnb } from "cnbuilder";
 import { alertTypes, lightText, darkText } from "./Alert.levers";
 import { DismissButton } from "../DismissButton/DismissButton";
+import "./alert.css";
 
 /**
  * Alert Component.
--- a/src/Alert/Alert.levers.js
+++ b/src/Alert/Alert.levers.js
@@ -6,10 +6,10 @@ export const alertTypes = ["info", "warning", "error", "success"];
 /**
  * Styles for light text
  */
-export const lightText = "su-text-white hover:su-link-no-underline";
+export const lightText = "su-text-white su-alert-body-link-light";
 
 /**
  * Styles for dark text
  */
 export const darkText =
-  "su-text-black su-link-black-true hover:su-link-no-underline";
+  "su-text-black su-link-black-true su-alert-body-link-dark";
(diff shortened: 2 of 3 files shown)


romakravchuk:
Hey @sherakama added styles as css file, tried to add like css module but to do that we need to configure storybook postcss addon in config file(main.js). So i make it simple and just import css file directly to js file

sherakama:
@romakravchuk I think I may have lead you astray. Please move this work directly into the saa_alumni repo.

sherakama (changes requested):
Please move to ~https://github.com/SU-SWS/saa_alumni~ https://github.com/SU-SWS/ood_online_giving

romakravchuk:
oh, and to online_giving as well??  i need this changes there also

sherakama:
> @sherakama oh, and to online_giving as well?? i need this changes there also

My mistake again. Just the online giving site. Not the SAA Alumni site at this time. Thanks for your patience.

sherakama:
> no worries, 😉
> i have [task](https://stanfordits.atlassian.net/browse/ADAPT-2066) to build the alert for alumni, so i should ignore this changes for alumni

Let me check on that. I'm now not sure if the link styles are different or not on those two sites.

yvonnetangsu:
It might be slightly easier on SAA Alumni site, since we have actually have TW installed there with a custom tailwind.config.js. We likely would need custom TW utilities/components to change the link background color in the WYSIWYG text, but probably no need for CSS modules.

sherakama:
Ok, I got clarity from our UX team, in that yes, we do want these changes in this package. Let's noodle on this a bit about the approach because this is definitely a gap in TailwindCSS's approach to styling. Please continue with the implementation of the banners for the two sites minus this link hover state.

sherakama:
@yvonnetangsu & @romakravchuk 

Ok, after deliberation on this we are going to go full circle. I'd like this moved back into each of the two other projects. I'd rather not introduce css modules to this library at this point as that will assume that the consumer has something like webpack with a css-loader to handle it. Please close this PR and move the work into each of the other two projects (SAA Alumni and OOD Online Giving). We will create a ticket to revisit this issue in Decanter-React later.

yvonnetangsu:
@sherakama Sounds good to me!

PR #56, opened 2021-08-03
ADAPT-2102: Hero component

rebeccahongsf:
# READY FOR REVIEW

# Summary
- Add Hero component

# Needed By (Date)
- tbd

# Urgency
- low

# Steps to Test

1. Checkout this branch
2. Spin up local dev to review the Hero or see on [Netlify Preview](https://deploy-preview-56--decanter-react.netlify.app/?path=/story/composite-hero--default)

- [ADAPT-2102](https://stanfordits.atlassian.net/browse/ADAPT-2102)

# See Also


rebeccahongsf:
Decanter v6 Hero contains [custom heights](https://github.com/SU-SWS/decanter/blob/v6/core/src/scss/components/hero/_hero.scss#L3-L11) at specific breakpoints. I've current set it to `su-max-h-500` (i.e. max-height: 50rem).

Is this something we should incorporate within our [maxHeight file](https://github.com/SU-SWS/decanter/blob/v7/src/plugins/theme/maxHeight.js) in Decanter v7? 🤔

yvonnetangsu:
> Decanter v6 Hero contains [custom heights](https://github.com/SU-SWS/decanter/blob/v6/core/src/scss/components/hero/_hero.scss#L3-L11) at specific breakpoints. I've current set it to `su-max-h-500` (i.e. max-height: 50rem).
> 
> Is this something we should incorporate within our [maxHeight file](https://github.com/SU-SWS/decanter/blob/v7/src/plugins/theme/maxHeight.js) in Decanter v7? 🤔

Oh interesting....those are very specific max heights. Hmmmm let's go with the closest value for now, since we aren't going for an exact 1:1 match with Decanter 6 currently. If we absolutely need those values later, we can always add them 😄

src/Hero/Hero.js
```diff
@@ -0,0 +1,63 @@
...
+import { dcnb } from "cnbuilder";
+
+/**
+ * Hero Component
+ *
+ * HTML card element
+ */
+
+const HeroRoot = ({
+  bgImage: { src } = {},
```
sherakama, line 12: If you are not using any other values than src why make this parameter an object?

sherakama (approved):
Other than my one question about the bgImage prop being an object vs a string this is looking great and is gtg.

yvonnetangsu:
This looks great! I was thinking though, the one with the caption seems like it belongs in a different family. I think we're going to need a component with image and caption (kind of like the CaptionMedia component we have on SAA). Think we could move the Hero with the small caption into a a different component, then we could add other options later (eg, caption alignment) 😄

yvonnetangsu (changes requested):
Sorry for taking so long to get back to this! 

1. Looks like currently it might breakdown for mobile if the text gets long. Please see if you could make the image height expand to accommodate the amount of text 😄  
Screen Shot 2021-09-02 at 1 48 48 AM

2. We could probably use larger font sizes for the description and maybe even the superheadline. I suggest talking to Kerri (since she's the design gatekeeper of Decanter) to see if she has font size suggestions.

Looking great! Thank you for the work 🙏🏼

rebeccahongsf:
Mini Update: 
1. Text issue for the Custom Gradient with text overlay should now be resolved for mobile:
image

2. Checked in with Kerri — will update PR with rec font-sizes once I hear back!

yvonnetangsu:
Yay - thank you for the revision, @rebeccahongsf ! ❤️

rebeccahongsf:
I think we're ready for another round of reviews @yvonnetangsu :D Thank you for your patience and feedback!

yvonnetangsu:
@rebeccahongsf Looks like default hero mobile view got a little chopped off on mobile at the bottom 😄 
Screen Shot 2021-09-21 at 11 59 06 AM

rebeccahongsf:
@yvonnetangsu, now it should be g2g. Thanks again!

yvonnetangsu:
@rebeccahongsf One more 😄 
I would probably take away the max height on that 2nd variant:
Currently with max height 500:
Screen Shot 2021-09-21 at 12 49 42 PM

After
Screen Shot 2021-09-21 at 12 49 58 PM
:

PR #59, opened 2021-08-16
ADAPT-3101 Refactor CtaLink (work-in-progress)

rebeccahongsf:
# NOT READY FOR REVIEW

# Summary
- Refactor CtaLink (alternative solution) — Additional testing and research needed to see how we would be able to pass or bind an Element/Component (i.e. Gatsby Link, Storyblok Link, etc).

# Needed By (Date)

# Urgency
- How critical is this PR?

# Steps to Test

1. Do this
1. Then this
2. Then this

- ADAPT-3101

# See Also

### DIFF
--- a/src/CtaLink/CtaLink.js
+++ b/src/CtaLink/CtaLink.js
@@ -15,12 +15,29 @@ import { SrOnlyText } from "../SrOnlyText/SrOnlyText";
  */
 export const CtaLink = React.forwardRef(
   (
-    { className, text, srText, color, icon, iconProps, animate, ...props },
+    {
+      className,
+      element,
+      text,
+      srText,
+      color,
+      icon,
+      iconProps,
+      animate,
+      ...props
+    },
     ref
   ) => {
     // Defaults & Variables.
     // ---------------------------------------------------------------------------
     const levers = {};
+    let Element = "a";
+
+    // Determine whether a or element tag should be used based on children props
+    // ---------------------------------------------------------------------------
+    if (element) {
+      Element = { element };
+    }
 
     // Levers
     // ---------------------------------------------------------------------------
@@ -63,7 +80,7 @@ export const CtaLink = React.forwardRef(
     const { className: iconClasses, ...iProps } = iconProps || {};
 
     return (
-      <a
+      <Element
         className={dcnb(
           "su-cta-link su-text-19 md:su-text-20 su-block su-w-fit su-no-underline hover:su-underline focus:su-underline su-group su-transition-colors",
           levers.color,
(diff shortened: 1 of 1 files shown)


src/CtaLink/CtaLink.js
```diff
@@ -15,12 +15,29 @@ import { SrOnlyText } from "../SrOnlyText/SrOnlyText";
...
     // Defaults & Variables.
     // ---------------------------------------------------------------------------
     const levers = {};
+    let Element = "a";
+
+    // Determine whether a or element tag should be used based on children props
+    // ---------------------------------------------------------------------------
+    if (element) {
+      Element = { element };
+    }
```
sherakama, line 40: Instead of changing the element, when a child element is present, use what was passed in as the root element returned but modifying it first by cloning it with your added config (props)

Re: https://reactjs.org/docs/react-api.html#cloneelement

```
... 
if (children) {
  let Element = React.cloneElement(children, {your props here})
}

Comment (sherakama) [reply src/CtaLink/CtaLink.js:40]: Some experimenting with this will be required around the children values.

Comment (rebeccahongsf) [reply src/CtaLink/CtaLink.js:40]: Did a few codesandbox tests to understand how `React.cloneElement()` works then tried implementing it within the CtaLink. Mmm, I don't believe we can use the method in a functional component. We'll need to rewrite the CtaLink as a class component which could conflict with `React.forwardRef`. 🤔  We might need to go back to the drawing board for an alternative solution.

Just to confirm — our end goal here is to be able to pass a component (i.e. Gatsby <Link>, StoryBlok <SbLink>) into our CtaLink component so that the Icon would render within the following html structure:
 
```
<a href="/">
     Cta Link <svg /> 
     // where the <svg /> is the Icon from the <Icon /> component
</a>
```
instead of the implementation written in #58, where it would render:

```
<span>
     <a href="/">
          Cta Link
     </a>
     <svg /> 
     // where the <svg /> is the Icon from the <Icon /> component
</span>
``` 
right? 😃

yvonnetangsu: I'll let Shea answer the implementation question, but want to come in here and say that I like your first html structure better. It's more semantic, and works better with TW classes, e.g., `group-hover`, `group-focus`. For the 2nd one, to style the icon focus states, we'll have to use a `group-focus-within` class, which would require us to turn on more variants since `group-focus-within` isn't generated by default decanter core v7 😄  In either case, we'll need to be able to pass classnames and ref to the link also.

rebeccahongsf: @sherakama a penny for your thoughts on the implementation and what we're trying to achieve on the HTML structure side of things for the CTA Links? 🤔 

I'll revisit this in the upcoming sprint to see if there's another approach we can take as well.

rebeccahongsf:
PR clean up! — Closing out since this is no longer needed.

PR #60, opened 2021-08-16
ADAPT-3019 Update Decanter to 7.0.0-beta.0

rebeccahongsf:
# READY FOR REVIEW

# Summary
- Update Decanter to 7.0.0-beta.0
This has been tested in the Make a Gift repo here: https://github.com/SU-SWS/ood_online_giving/pull/84
Aside from the checkbox and radio button label change (missing `vertical-align: top`) everything appears as expected.

# Needed By (Date)

# Urgency
- How critical is this PR?

# Steps to Test
1. Checkout this pull request
2. Review code
3. Test out https://github.com/SU-SWS/ood_online_giving/pull/84 and see if there are any additional visual changes. 

- [ADAPT-3019](https://stanfordits.atlassian.net/browse/ADAPT-3019)

# See Also


yvonnetangsu (commented):
Sorry for taking so long to get back to you on this, but the dev chat on Monday was helpful, and thanks to you adding the out of the box form elements here, it's clearer for us to see what they currently look like with just Decanter v7 base styles without the classnames. 

I only see one common issue with some button variants (any of the non red ones) and the dismiss buttons. I think due to the Decanter refactor, the hover/focus states for eg, the outline and ghost buttons, or the dismiss buttons, are now taking on the black background with white text style. On live and in the design, they actually should behave differently (see https://decanter-react.netlify.app/?path=/story/simple-dismiss-button--default for reference, the hover/focus bg is transparent). There could be multiple factors, one is that, we removed the "hocus" and "group-hocus" variants from Decanter v7 core to reduce bloat (and we normally use JIT mode anyway), so those variants that are used in DR need to be replaced with `hover:` and `focus:` classes separately instead of just `hocus:` for example (same for the `group-` ones). Also, specificity of CSS might have changed so extra class names might be needed.

With that said, I suspect if we end up adding JIT mode into this repo, then you wouldn't need to change things back and forth which probably saves time 🤔 😂 

Hmmm....decisions decisions.

sherakama:
@yvonnetangsu What is the pivot point for this to be merged? 

I see the one issue with the dismiss button to be cleaned up still too.

yvonnetangsu:
> @yvonnetangsu What is the pivot point for this to be merged?
> 
> I see the one issue with the dismiss button to be cleaned up still too.

Yeah, I think the Dismiss Buttons and the non-primary Button variants need to be tweaked, looks like the CTA Buttons have some hover/focus state issues as well
Screen Shot 2021-09-03 at 11 46 53 AM

If Nathan is working on getting JIT mode into this repo, would it make more sense to do these changes after?

sherakama (approved):
I merged in https://github.com/SU-SWS/decanter-react/pull/62 (JIT WORK) and re-packaged. This is G2G once https://github.com/SU-SWS/decanter-react/pull/62 has been merged.

rebeccahongsf:
Thanks @sherakama 😄 Seems like the [button](https://6140dd0a9648610acd3d0d87--decanter-react.netlify.app/?path=/story/simple-button--solid) issues have been resolved as well @yvonnetangsu but please do not hesitate to let me know if anything else appears out of place!

yvonnetangsu:
> Thanks @sherakama 😄 Seems like the [button](https://6140dd0a9648610acd3d0d87--decanter-react.netlify.app/?path=/story/simple-button--solid) issues have been resolved as well @yvonnetangsu but please do not hesitate to let me know if anything else appears out of place!

Yay! Nathan is making one final change to the JIT branch. After that, we'll remerge/repackage and then it should be good to go. Right now it's missing the logo styles from JIT.

rebeccahongsf:
Decanter has been updated within PR #62
See: https://github.com/SU-SWS/decanter-react/blob/main/package.json#L53
Closing this PR out 😄

PR #65, opened 2021-09-11
ADAPT-3096: Text Field

rebeccahongsf:
# READY FOR REVIEW

# Summary
- Text Field Component Proof of Concept: TextField should contain subcomponents that pull in InlineError and Label.
**Note:** Additional updates will need to be done for the InlineError icon(s) (along with any other components that use the Hero Icon package). 

# Needed By (Date)
- tbd

# Urgency
- medium

# Steps to Test

1. Pull this branch
2. Review locally at `http://localhost:6006/?path=/docs/form-elements-text-field--default` or view [Netlify preview](https://deploy-preview-65--decanter-react.netlify.app/?path=/story/form-elements-text-field--default)
3. Review code and evaluate the component(s) and how it is assembled.
Some things to consider while reviewing code:
- Do the components follow the atomic design methodology?
- How might we implement the component with react-hook-forms? 
- What adjustments can be made to better accept props from react-hook-forms?

- Decanter-React-Forms (DERF)

- [ADAPT-3096](https://stanfordits.atlassian.net/browse/ADAPT-3096)
cc: @sherakama @yvonnetangsu 

# See Also


src/InlineError/InlineError.js
```diff
@@ -0,0 +1,75 @@
...
+
+// Prop Types.
+// -----------------------------------------------------------------------------
+InlineError.propTypes = {
+  // CSS Classes.
+  className: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.array,
+    PropTypes.object,
+  ]),
```
mattanglin, line 54: I know this something that's been kind of standardized on here, but I feel like extending the `className` prop to anything other than it's native `string` value is a dangerous practice. Plus, I've already ran into it [throwing console warnings](https://github.com/SU-SWS/saa_alumni/pull/159#discussion_r705602877) in the homesite project.

src/InlineError/InlineError.js
```diff
@@ -0,0 +1,75 @@
...
+  text: PropTypes.string,
+
+  // Icon name
+  icon: PropTypes.string,
+
+  // Icon Type
+  iconType: PropTypes.string,
+
+  // Is there an error message?
+  isError: PropTypes.bool,
```
mattanglin, line 66: Is this necessary? Couldn't we just change `text` to be an optional string property named `error` or `errorText` and render based on its existence?

src/InlineError/InlineError.js
```diff
@@ -0,0 +1,75 @@
...
+        type={iconType}
+        aria-hidden="true"
+        className="su-text-digital-red su-inline su-mr-3"
+        {...iconProps}
+      />
+    );
+  }
+
+  return (
+    <>
```
mattanglin, line 33: This fragment wrapper is unnecessary since we're returning a single node conditionally. You can optionally return `null` to render nothing and keep the (virtual) DOM a little cleaner.

src/TextField/TextField.stories.js
```diff
@@ -0,0 +1,52 @@
...
+  <TextField>
+    <TextField.Label text={labelText} helpText={helpText} id={id}>
+      <input
+        {...props}
+        className="su-input su-border su-border-black su-border-b-2 su-rounded su-pt-19 su-pb-20 su-px-19 su-w-full"
+        id={id}
+      />
+      <TextField.InlineError text={errorText} isError={isError} />
+    </TextField.Label>
+  </TextField>
```
mattanglin, line 38: I'm still struggling with why we're creating a component library and requiring consumers to do this much integration work to use it. If we're creating a library to speed up our own processes, I'd (personally) much rather import a single component and pass props to it than have to compose it this way every time (or more likely just recreate this exact code as a component in my codebase for reuse).

rebeccahongsf: Thanks for the swift review and feedback, @mattanglin! 😃 The way it's currently built is similar to how we built the Poster component (inspired by [HeadlessUI](https://headlessui.dev/)). 

However, I can definitely see how we might restructure the TextField component to be a singular component since that reduces the need to recreate the component on the project level! I threw this alternative TextField component poc that I believe embodies the idea you had in mind: https://github.com/SU-SWS/decanter-react/pull/68

We can chat more about the direction we want to go on this with the team during ADAPT Dev Chat today.

src/InlineValid/InlineValid.js
```diff
@@ -0,0 +1,66 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { dcnb } from "cnbuilder";
+import Icon from "react-hero-icon";
+
+export const InlineValid = ({ className, validText, icon, iconType }) => {
```
rebeccahongsf, line 6: Does it make sense to create an InlineValid component to display validation messages? Or is there an elegant way we can reuse the InlineError component to toggle between Valid and Error messages/styles 🤔 

I imagine there's a way to reuse and recycle — I'll poke around RHF documentation to see how these messages are typically absorbed by the component.

mattanglin: You could combine the two, but I think it's fine to keep them separate since they serve two different, mutually exclusive purposes despite the fact that they generally occupy the same space in the DOM structure.

src/InlineError/InlineError.js
sherakama: > I feel like extending the className prop to anything other than it's native string value is a dangerous practice

The trade here is the ability to unset styles that we have added to the outermost wrapper by passing in an object of falsy values for the things we don't want through to dcnb(). This can allow us to keep the list of outputted classnames cleaner instead of appending opposing classnames to the end of the class attribute.

eg: If we put `su-rs-4` but my project wanted no spacing we could pass in `{su-rs-4: false}` to remove it from the output instead of adding `su-rs-0`.

src/TextField/TextField.stories.js
sherakama: > I'm still struggling with why we're creating a component library and requiring consumers to do this much integration work to use it.

It is unusual for sure but so is the TW component space to me. Copy and pasting HTML markup around to "share" components is strange and doesn't _feel_ right. That said, _some assembly required_ is where headlessUI and we landed. Our scope for this component library is not just for one site, but rather, for a suite of sites that are not identical but use a lot of the same things. The ability to add customizations and themeability to the out-of-the-box elements is a highly needed feature. The DX for that can go back and forth but I feel that the copy and paste culture of tailwind met a nice intersection with React in the HeadlessUI library. Coupling our component library with Storybook documentation should give the end developer all they need to get up and going as well as access to each and every element to theme, style, swap out, or append to. It is a little extra to have to copy and paste from a documentation website instead of just import and send through props but those familiar with the tailwind ecosystem should feel at home.

sherakama: Happy to continue this discussion and ways to make the DX better in any of our meetings and here to.

sherakama (approved):
Thanks for everyone's feedback on this. I vote for merging this down and starting to put it through its paces in real project work as we can always change things as we go.

yvonnetangsu:
> Thanks for everyone's feedback on this. I vote for merging this down and starting to put it through its paces in real project work as we can always change things as we go.

Do it 👍🏼  Merge away 😄  Let's get it in and we can tweak as needed.

PR #74, opened 2021-09-24
ADAPT-3307: Autocomplete component (contains ADAPT-3308)

Lyafel:
# Summary
- ADAPT-3307 and ADAPT-3308. Autocomplete with one and multiple options with the ability to use only the provided options or add new ones. 

# Needed By (Date)

# Urgency
- How critical is this PR?

# Steps to Test

1. Do this
1. Then this
2. Then this

- JIRA ticket
- Other PRs

# See Also


Lyafel:
@yvonnetangsu
- I just realized that it looks like it was intended that the design would be consistent between all inputs of the kind like text and select but the select looks differently, I'll ask Bhavika in the ticket, still creating the pr because I fell like I started going in circles and would want so see what you think about this solution.
- Looks like ADAPT-3309 asks for a mobile fallback, does this actually need anything special or just reuse the select from ADAPT-3126?

rebeccahongsf (approved):
Redeployed the Netlify preview and manually resolved code-climate. Everything looks great to me 😸 I'll defer to @yvonnetangsu for final review and merge.

Thank you @Lyafel!

yvonnetangsu:
@Lyafel These are looking great. Would like a couple changes:

- Please make the whole chip a button instead of just the "X" being the button
- disabled state shouldn't have hover/focus/active styles

Thanks!

Lyafel:
@yvonnetangsu Done

sherakama (changes requested):
@Lyafel could you update to MUI version 5 for this package is it is now out and ready for use.

package.json
```diff
@@ -66,6 +66,8 @@
     "@babel/preset-env": "^7.12.17",
     "@babel/preset-react": "^7.12.13",
     "@fullhuman/postcss-purgecss": "^3.0.0",
+    "@material-ui/core": "^4.12.3",
```
sherakama, line 69: Is it possible to use version 5?

tailwind.config.js
```diff
@@ -63,6 +64,18 @@ utilityFiles.forEach((file) => {
...
+// eslint-disable-next-line no-shadow
+const autocomplete = plugin(({ addComponents, theme }) =>
+  addComponents({
+    '.autocomplete-option[data-focus="true"]': {
+      backgroundColor: theme('colors.digital-blue.DEFAULT'),
+      color: theme('colors.white'),
+    },
+  })
+);
+plugins = plugins.concat(autocomplete);
```
sherakama, line 77: This could be a problem as we are not requiring individuals that consume this package to include the TailWindCSS confirmation from this package. Can you not get at the options to style?

src/Autocomplete/Autocomplete.stories.js
```diff
@@ -0,0 +1,101 @@
+import React from 'react';
+import { AutocompleteField } from './Autocomplete';
+
+export default {
+  title: 'Composite/Autocomplete',
```
sherakama, line 5:
```suggestion
  title: 'Forms/Autocomplete',
```

src/Autocomplete/Autocomplete.stories.js
```diff
@@ -0,0 +1,101 @@
...
+const AutocompleteTemplate = ({ ...rest }) => (
+  <AutocompleteField options={options} {...rest} />
+  // // Or using with react-hook-form:
+  // <Controller
+  //   name="autocompleteField"
+  //   control={control}
+  //   render={({ field }) => {
+  //     return <AutocompleteField options={options} {...rest} {...field} />;
+  //   }}
+  // />
```
sherakama, line 63: This would be better as documentation on the storybook output instead of an inline code comment. Please remove the commented out code or move it to the storybook documentation for this component.

src/Autocomplete/Autocomplete.js
```diff
@@ -0,0 +1,218 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { createFilterOptions } from '@material-ui/lab/Autocomplete';
+import useAutocomplete from '@material-ui/lab/useAutocomplete';
+import Icon from 'react-hero-icon';
```
sherakama, line 5: @rebeccahongsf This will need an update once the new icon package is in.

tailwind.config.js
sherakama: Please remove this custom TWComponent and instead use the `componentsProps` option on the MUI components to inject your class names into the nested components.

sherakama: https://mui.com/customization/unstyled-components/

src/Autocomplete/Autocomplete.js
```diff
@@ -0,0 +1,218 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { createFilterOptions } from '@material-ui/lab/Autocomplete';
+import useAutocomplete from '@material-ui/lab/useAutocomplete';
```
sherakama, line 4: Please use the unstyled component option as we would rather not fight or have an intersection with CSS in JS and TailwindCSS

https://mui.com/customization/unstyled-components/

package.json
Lyafel: The issue is the react version, mui v 5 it requires 17 and we're on 16.

sherakama: Following up from our slack message. Yes, please update to React version 17. This is in flight everywhere.

tailwind.config.js
Lyafel: I removed it from the tailwind config, hovewer componentProps didn't work either for my case, used a different solution inside the component

package.json
```diff
@@ -50,6 +50,8 @@
   "homepage": "https://github.com/SU-SWS/decanter-react#readme",
   "dependencies": {
     "@heroicons/react": "^1.0.4",
+    "@mui/core": "^5.0.0-alpha.49",
+    "@mui/material": "^5.0.2",
```
sherakama, line 54: Please remove this package.

Lyafel: I'm using this for styling the focused item in the dropdown, '@mui/material/styles';. They have some custom magic going on that doesn't work with focus: tailwind styles, so I can apply hover styles but for focus I have to target their specific class and this seems to be mui's native way of doing things

sherakama:
@Lyafel This is now on hold. Please wait until further notice to where this work is going. We are going to be retiring this project and moving the work directly into projects.

sherakama:
Please move this work directly down and into the my account project.

PR #79, opened 2021-09-30
ADAPT-3130  MUI v5 Info Bubble/ToolTip

rebeccahongsf:
# NOT READY FOR REVIEW

# Summary
- Create InfoBubble component using the MUI v5 Tooltip component

# Needed By (Date)
- tbd

# Urgency
- medium

# Steps to Test

1. Pull this branch
2. Review and compare to figma mocks
- InfoBubble: `/story/form-elements-information-bubble--default` or [Netlify Preview](https://deploy-preview-79--decanter-react.netlify.app/?path=/story/form-elements-information-bubble--default)
- ToolTip: `/story/form-elements-tool-tip--default` or [Netlify Preview](https://deploy-preview-79--decanter-react.netlify.app/?path=/story/form-elements-tool-tip--default)
3. Check code for anything funky 🤔 

- [ADAPT-3130](https://stanfordits.atlassian.net/browse/ADAPT-3130)
- [ADAPT-3821](https://stanfordits.atlassian.net/browse/ADAPT-3821)

# See Also


src/InfoBubble/InfoBubble.js
```diff
@@ -0,0 +1,86 @@
...
+  };
+
+  const handleTooltipOpen = () => {
+    setOpen(true);
+  };
+
+  const bubbleStyles = {
+    tooltip: dcnb(
+      className,
+      '!su-px-18 !su-pb-18 !su-pt-12 !su-sans !su-text-14 !su-text-black !su-bg-black-10 !su-color-black-10 !su-shadow !su-max-w-[25ch]'
```
rebeccahongsf, line 29: Our Decanter TW classes seem to be overridden by the MUI styles without the TW important prefix. 🤔

src/InfoBubble/InfoBubble.js
```diff
@@ -0,0 +1,86 @@
...
+        }}
+        onClose={handleTooltipClose}
+        open={open}
+        disableFocusListener
+        disableHoverListener
+        disableTouchListener
+        title={text}
+        placement={alignment}
+        classes={bubbleStyles}
+      >
```
rebeccahongsf, line 49: I essentially copy-pasta (and tested!) this portion from [MUI Tooltip component with Triggers](https://mui.com/components/tooltips/#triggers) which seems to get the user interaction the way we want it to be 😃

src/InfoBubble/InfoBubble.js
```diff
@@ -0,0 +1,86 @@
...
+        PopperProps={{
+          disablePortal: true,
+        }}
+        onClose={handleTooltipClose}
+        open={open}
+        disableFocusListener
+        disableHoverListener
+        disableTouchListener
+        title={text}
+        placement={alignment}
```
rebeccahongsf, line 47: For some reason, I couldn't seem to get the levers (e.g. `const placement = alignOptions[alignment];` to work without encountering an error. For now I have the alignment props passed straight into placement and the alignOptions controls available for the user to select their preferred alignment in the Storybook UI. 

Do note that the MUI Tooltip is preconfigured so that it will not cut outside of the screen (e.g. if you set the alignment to `top`, it will display the Tooltip on the `bottom`. Cool stuff 😎

mattanglin: I believe mui does a lot of element level styling in addition to generating style classes that have more specificity than a single class selector (tbh, most react component libs do) so we're likely going to run up against this a lot when trying to apply tailwind classes 🦐

mattanglin: > I believe mui does a lot of element level styling in addition to generating style classes that have more specificity than a single class selector (tbh, most react component libs do) so we're likely going to run up against this a lot when trying to apply tailwind classes 🦐

That was supposed to be a `:shrug:`, but it autocompleted to `:shrimp:` and so there it is.

mattanglin: > > I believe mui does a lot of element level styling in addition to generating style classes that have more specificity than a single class selector (tbh, most react component libs do) so we're likely going to run up against this a lot when trying to apply tailwind classes 🦐
> 
> That was supposed to be a `:shrug:`, but it autocompleted to `:shrimp:` and so there it is.

Also this was in reference to the [other comment you made](https://github.com/SU-SWS/decanter-react/pull/79#discussion_r720544853). It's definitely the end of the week... 🤦🏼

yvonnetangsu: I was wondering about the 🦐 🤷🏼‍♀️ 😂 

And on the styling/!important TW classes topic:

So I haven't looked into the Unstyled MUI components. Are all components available with the Unstyled version? I wonder what's the effort is to style the Unstyled (if it's available) vs overriding the Styled ones.

package.json
```diff
@@ -49,7 +49,10 @@
   },
   "homepage": "https://github.com/SU-SWS/decanter-react#readme",
   "dependencies": {
+    "@emotion/react": "^11.4.1",
```
mattanglin, line 52: So here's where things start getting interesting (read convoluted); Apparently mui@v5 is now using emotion 👩🏼‍🎤  (which is a fantastic css-in-js solution if that's your thing) so we're adding that to our bundle now, too. Something to further discuss in dev talk...

src/InfoBubble/InfoBubble.js
mattanglin: I _believe_ that mui unfortunately only offers a subset of their components in unstyled versions and that `Tooltip` is not in that subet. Is that correct @rebeccahongsf ?

src/ToolTip/ToolTip.js
```diff
@@ -0,0 +1,58 @@
...
+
+  // CSS Classes.
+  className: PropTypes.string,
+
+  // Children
+  children: PropTypes.oneOfType([
+    PropTypes.node,
+    PropTypes.element,
+    PropTypes.string,
+  ]),
```
mattanglin, line 48: Just as an FYI, `PropTypes.element` and `PropTypes.string` are unnecessary here as they are [already encompassed by `PropTypes.node`](https://reactjs.org/docs/typechecking-with-proptypes.html#proptypes). The most accurate typing for `children` using `prop-types` would be the following:
```
children: PropTypes.oneOfType([
  PropTypes.node,
  PropTypes.arrayOf(PropTypes.node),
]),
```
The reasoning behind this being that sibling children are represented as ordered arrays. e.g.
```
<ParentComponent>
  <ChildA />
  <ChildB />
</ParentComponent>
// ParentComponent.props.children = [ChildA, ChildB]
```

mattanglin: hmmm... It looks like `PropTypes.node` might even encompass arrays now as well, so you might be able to just use that. 🤔

src/InfoBubble/InfoBubble.js
rebeccahongsf: That is correct @mattanglin! It seems like only the Autocomplete and Modal/Dialog are available unstyled in @mui/core ☹️

yvonnetangsu: Yeah....I suspected that's the case. A bit unfortunate but 🤷🏼‍♀️

sherakama:
@rebeccahongsf This is now on hold. Please wait until further notice to where this work is going. We are going to be retiring this project and moving the work directly into projects.

rebeccahongsf:
PR clean up! — Closing out since this is no longer needed.

